PR #4, opened 2016-09-25
Fix scripts for instances which have been upgraded from Owncloud

JimboJoe:
Instances upgraded from owncloud keep a database named "owncloud", and current backup/restore scripts won't work with that setting, as they are looking for a database named after the app.

This patch adapts scripts to look for database name in Nextcloud config file.

Backup/restore/remove has been tested on a "pure" Nextcloud instance, and an instance upgraded from Owncloud. Upgrade script hasn't been tested.

The fix looks ugly:
- a cleaner way could be to store the dbname in the YunoHost app settings during install/upgrade, but that would need to think about people that have already done the merge (thus not storing a YunoHost app setting)
- the code could be factorized in _common.sh but that would only work for upgrade/remove. I couldn't find a way to access the _common.sh script in backup/restore scripts.

### DIFF
--- a/scripts/backup
+++ b/scripts/backup
@@ -6,8 +6,8 @@ set -eu
 app=$YNH_APP_INSTANCE_NAME
 
 # Set app specific variables
-dbname=$app
-dbuser=$app
+dbname=`sudo cat /var/www/$app/config/config.php  |grep dbname|sed "s|.*=> '\(.*\)'.*|\1|g"`
+dbuser=$dbname
 
 # Source app helpers
 source /usr/share/yunohost/helpers
--- a/scripts/remove
+++ b/scripts/remove
@@ -5,8 +5,8 @@ source ./_common.sh
 
 # Set app specific variables
 app=$APPNAME
-dbname=$app
-dbuser=$app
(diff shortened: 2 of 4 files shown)


jeromelebleu:
Many thanks! I didn't thought and tested that... However, I would prefer to *rename* the database name and user while migrating from ownCloud to be consistent. It will also prevent to have to take care about that later in the scripts. I'll try to put that in the `upgrade` script so that users who already migrated their ownCloud app will also benefit from that.

JimboJoe:
OK, that seems better if you don't mind the effort!
Another nice effort would be to propose to change the app URL, because many instances use the default URL (i.e. "/owncloud"). But I know that means messing with Nextcloud internals...

jeromelebleu:
Done with https://github.com/YunoHost-Apps/nextcloud_ynh/commit/842080755658ecfae81a66bde822d2d04bc1b370. Thanks anyway for your PR, it was morevover helpful to retrieve the `dbname` value from `config.php`! Could you test it to be sure that it's working for you too please?

JimboJoe:
Tested and approved, thanks!

PR #24, opened 2017-03-12
Apply Nextcloud #237 fix

JimboJoe:
This is a recommended fix for Nginx configuration. More information [here](https://github.com/nextcloud/server/issues/237#issuecomment-229923262).


maniackcrudelis:
Euh yes why not. But why?
We don't have any 503 error?

Issue #26, opened 2017-03-19
Missing YunoHost tile


JimboJoe:
The YunoHost tile is missing in this application.
Unfortunately, even after applying latest fixes for missing tiles (see [this bug](https://dev.yunohost.org/issues/285)), it still doesn't show up.
The root cause seems to be the CSP nonce rules implemented in Nextcloud 11, that prevent the `ynhpanel.js` to be loaded by the browser.
I see no fix/workaround at this time, unless creating a dedicated Nextcloud app that would link back to the YunoHost menu.

Issue #18, opened 2017-01-23
Content Security - Scripts


pavanagrawal123:
Hi,

I just installed this app into my Yunohost, however on launching the software it's not working because of the content security on the JS scripts. Is this going to be fixed on future releases or do i need to make a configuration change?

juliencognito:
Hello
Same problem on my side; with firefox or chromium. The console view of the firefox's development tools tells me about 'unsafe-eval'.
The files are not displayed until I voluntarily click on 'files'; As long as I am connected, the problem is resolved.
As for the Ynh_Panel, it never appears.

I encounter this problem with NextCloud, not with other applications & not on https://demo.nextcloud.com

Thanks for attention,

Julien

Shnoulle:
Same issue : Content Security Policy on 1st load, reload : all;files are loaded.

My nextcloud are on a subdirectory of my ynh login page.

jellium:
I encounter the same issue.
Below the first of the many errors reported by the browser debug tools:

`Content Security Policy: The page’s settings blocked the loading of a resource at self (“script-src 'nonce-S0VxMFBpTEFvcCt4OUIvbXNRMFk0ZWRjUW1zMm9Fa1NJMjRpamhKeVFNOD06UkhMTlVGU3h4TmJWbzJxbDVENTVsNjlwYlQ1NjhSaGdSd3BOK1haRUthRT0=' 'unsafe-eval'”).`

which points to the following line in the source:

`<script nonce="NHNnSzJkYzltV0I3VHpqWldGd00rMHczbkV5MVRSQmVlZjMrMW43RUdkMD06MTc0K2phTm43d3NNWkZhRGQyMGpqQ1pScWlqOUpVWWFRSWpLbmhXSWE2Zz0=" type="text/javascript">`

After refreshing the page once, the problem does not occur and line 55 becomes:

`<script nonce="SUdmWW9iWi9YTlc1U29reVJqMTlXcDZIb3FFRldqTUVzbmJYM1JiS2J5Zz06VEYraHo4QU9PcHpkSGZ4eEV3NGNMTmF5amZSSkMySjIxaEs0cW5MOEJrWT0=" type="text/javascript">`

This behavior is also encountered in the "private browsing" mode of Firefox as well.

Shnoulle:
Tracking done error source (for my error : reload load the scripts)

[...]

PR #28, opened 2017-04-01
fix Missing YunoHost tile (#26) and blank page at login due to CSP scripts rules (#18)

ianbogda:
add more headers to change nextcloud CSP.
yuonohost tile appears again

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -20,6 +20,7 @@ location ^~ #LOCATION# {
   add_header X-Robots-Tag none;
   add_header X-Download-Options noopen;
   add_header X-Permitted-Cross-Domain-Policies none;
+  more_set_headers Content-Security-Policy "default-src 'self' 'unsafe-eval' data:;";
 
   # Set max upload size
   client_max_body_size 10G;
@@ -82,6 +83,7 @@ location ^~ #LOCATION# {
     add_header X-Robots-Tag none;
     add_header X-Download-Options noopen;
     add_header X-Permitted-Cross-Domain-Policies none;
+    more_set_headers Content-Security-Policy "default-src 'self' 'unsafe-eval' data:;";
     # Optional: Don't log access to assets
     access_log off;
   }
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -20,7 +20,8 @@ location ^~ #LOCATION# {
   add_header X-Robots-Tag none;
   add_header X-Download-Options noopen;
   add_header X-Permitted-Cross-Domain-Policies none;
-  more_set_headers Content-Security-Policy "default-src 'self' 'unsafe-eval' data:;";
+  # Add data: to allow /ynhpanel.css to be load due to image on data:base64
+  more_set_headers Content-Security-Policy "default-src data:;";
 
   # Set max upload size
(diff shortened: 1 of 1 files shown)


JimboJoe:
Thanks @ianbogda for this pull request!
I tested it and it indeed works :-)
Could you please explain the rational behind this modification?
I don't know much about CSP, but does this change disable Nextcloud CSP rules altogether (thus possibly degrading security)?

P.S. : It does also seem to fix the blank page after first login through SSO (CSP rules problems as well), which wasn't linked to YunoHost tile, so it does emphasize my previous question :-)

ianbogda:
Hye @JimboJoe,
after more investigations, rules from nextCloud and tests :)
L23```more_set_headers Content-Security-Policy "default-src  data:;";``` is enough due to **/ynhpanel.css** where yunohost image tile and fonts are **data:base64**.

There is no SP leaks in this case.

I'll send rectification in this way.

conf/nginx.conf
```diff
@@ -20,6 +20,8 @@ location ^~ #LOCATION# {
   add_header X-Robots-Tag none;
   add_header X-Download-Options noopen;
   add_header X-Permitted-Cross-Domain-Policies none;
+  # Add data: to allow /ynhpanel.css to be load due to image on data:base64
```
JimboJoe: typos: ...load**ed** due to **inline** image...

conf/nginx.conf
```diff
@@ -20,6 +20,8 @@ location ^~ #LOCATION# {
...
+  more_set_headers Content-Security-Policy "default-src data:;";
```
JimboJoe: here wouldn't it be safer by adding `'self'` directive: `default-src 'self' data:;`?

loicdeter: Bye default, nextCloud refuse data: for css images.

the nginx.conf will just allow from yuonohost instance, so there is no security leak.

JimboJoe: In my understanding, CSP rules are applied by the web browser. Nginx is only concerned because we're adding that precise header.
My question was oriented towards what you're saying: allowing only from YunoHost instance. Wouldn't that require the `'self'` directive?

opi: wrong syntax, so this CSP (`default-src data:;`) is never applied.

But due to `more_set_headers` syntax, this line completely remove the current CSP header. See https://github.com/openresty/headers-more-nginx-module#more_set_headers

A correct syntax would be:
```
more_set_headers "Content-Security-Policy: default-src data:;";
``` 

but having `default-src data:;` as CSP broke everything !!  To see what current CSP is applied, use `curl -I https://domain.com |grep '^Content-Security-Policy:'`

mokas01:
The solution in order not to ruin the CSP could be in SSOwat conf
https://dev.yunohost.org/issues/924
(?)

alexAubin:
(Bump?)

JimboJoe:
As @opi demonstrated, your PR unfortunately totally disables the CSP mechanism.
As another PR has been opened [here](https://github.com/YunoHost-Apps/nextcloud_ynh/pull/52) to solve these two problems, I'll close this one for now.
Thanks a lot anyway for your contribution! 👍

PR #57, opened 2017-08-18
[Fix] Yunohost Version (Fixes #56)

### LINKED ISSUE #56: Nextcloud installation fails on Yunohost 2.5
The nextcloud_ynh does not seem to check the version of yunohost, and always use the ynh_replace_string, which is new to yunohost 2.6.

The official image for raspberry pi is still 2.5, so the installation fails.

It can be solve by running on the command line a:

`apt-get update && apt-get dist-upgrade.`

However, it might be relevant to check the version and install accordingly.

Best regards

frju365:
*Micro decision*

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -20,7 +20,7 @@
         "mysql"
     ],
     "requirements": {
-        "yunohost": ">= 2.4.0"
+        "yunohost": ">= 2.6.0"
     },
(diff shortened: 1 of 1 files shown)


maniackcrudelis:
You can merge that as a micro decision ;)

PR #59, opened 2017-09-01
Big refactoring !

maniackcrudelis:
## Problem
- *Need a whole refactoring and a simplification !*
- *Owncloud migration seems to not work anymore*
- *We need to have a migration process for all forked apps*

## Solution
- *Package refactored to adjust to latest standards*
- *Sequential upgrading process rewrited*
- *Add a new helper ynh_handle_app_migration to managed more easily the migration of an app*

## PR Status
Could be reviewed and tested.  

## Validation
~*Medium decision*~
- [ ] ~**Complete test** :~
- [ ] ~**Upgrade previous version** :~
- [ ] ~**Code review** :~
- [ ] ~**Code review** :~
- [ ] ~**Approval (LGTM)** :~
- [ ] ~**Approval (LGTM)** :~
- [ ] ~**Approval (LGTM)** :~
- [ ] ~**CI succeeded** :~
- [ ] ~**CI succeeded** :~
~One public package check log is enough instead of two checks.  
When the PR is mark as ready to merge, you have to wait for 7 days before really merge it.~

Since this PR has been rerouted from master to testing. It passes to a minor decision.
- [X] **Upgrade previous version** : JimboJoe
- [X] **Code review** : JimboJoe
- [X] **Approval (LGTM)** : JimboJoe
- [X] **CI succeeded** : JimboJoe


scripts/_common.sh
```diff
@@ -53,222 +55,250 @@ exec_as() {
...
-ynh_store_file_checksum () {
-	local checksum_setting_name=checksum_${1//[\/ ]/_}	# Replace all '/' and ' ' by '_'
-	ynh_app_setting_set $app $checksum_setting_name $(sudo md5sum "$1" | cut -d' ' -f1)
-}
-
-# Verify the checksum and backup the file if it's different
-# This helper is primarily meant to allow to easily backup personalised/manually 
-# modified config files.
+# If you have some dependencies for your app, it's possible to change the fake debian package which manages them.
+# You have to fill the $dependencies variable, and then a new fake package will be created and installed,
```
JimboJoe: Many apps use `$pkg_dependencies`, shouldn't we use that name instead?

maniackcrudelis: As you want :)

scripts/backup
```diff
@@ -1,48 +1,86 @@
...
+ynh_backup "/etc/php5/fpm/conf.d/20-$app.ini"
 
-# Copy the app source files
-DESTDIR="/var/www/$app"
-ynh_backup "$DESTDIR" "www"
+#=================================================
+# BACKUP THE MYSQL DATABASE
+#=================================================
+
+ynh_mysql_dump_db "$db_name" > db.sql
```
JimboJoe: Here and in every other refactoring: is there anything special with `db.sql` file so that we assume if will be automatically backed-up...?
In my opinion, we lack at least a comment on that point...

maniackcrudelis: Because you make a dump in the current directory, it's not needed to copy this file a second time.
It will be backuped with all the other files in the current directory.

And if you try to use `ynh_backup` with this file, you'll have a error because this file already exist.

JimboJoe: OK, so I understand the current directory is implicitly backed-up. Maybe should we state that?

frju365:
Is this PR still alive ?

JimboJoe:
Yes, it's sill alive :-)
I've just resolved the conflict with the newer 12.0.2 release.

JimboJoe:
Tested migration from Nextcloud 9, 10, 11 and 12 OK 👍 
Migration from owncloud 9 wasn't smooth:
- the application is still named "owncloud"
- the path is still /owncloud (but I guess it's perfectly normal as we can't type a new path at upgrade)
- I got this log in `/tmp/owncloud_post_migration.sh.log` : 
```
/bin/bash: BASH_XTRACEFD: 7: invalid value for trace file descriptor
The SSOwat configuration has been generated
Remove the user owncloud
warning: commands will be executed using /bin/sh
job 2 at Sat Sep 23 07:19:00 2017
```

maniackcrudelis:
> the application is still named "owncloud"

Now, the label should change if it's according to the name of the app. So, if your app is named Owncloud or owncloud, it's becoming Nextcloud.

> the path is still /owncloud (but I guess it's perfectly normal as we can't type a new path at upgrade)

We should add a change_url script to do that. Not sure it should be in the upgrade process though.

> I got this log in /tmp/owncloud_post_migration.sh.log

It's the log of the postponed script. Should we put this log elsewhere ?

JimboJoe:
Label was changed to Nextcloud with the fix 👍 
So this PR will be merged to testing in 3 days.
I recommend to upgrade testing to the new NextCloud 12.0.3 before posting on the forum.

PR #68, opened 2017-10-06
Fix a lonely Roundcube comment

Bridouz:
- One comment said it was retrieving Roundcube tarball instead of Nextcloud one.


maniackcrudelis:
Thanks for this attention

However, this function, extract_nextcloud, is no longer used in the new version of the package.
As you can seen in the testing branch.
https://github.com/YunoHost-Apps/nextcloud_ynh/blob/testing/scripts/_common.sh

Thank you anyway

PR #96, opened 2018-03-08
Fix stretch

maniackcrudelis:
## Problem
- *Instal doesn't work on stretch*

## Solution
- *Add php-zip as a dependence.*

## PR Status
Could be reviewed and tested.

## Validation
---
*Minor decision*
- [x] **Upgrade previous version** : JimboJoe
- **CI succeeded** : [Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20fix_stretch%20(Official)/)  

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -3,7 +3,7 @@
 # COMMON VARIABLES
 #=================================================
 
-pkg_dependencies="php5-gd php5-json php5-intl php5-mcrypt php5-curl php5-apcu php5-redis php5-ldap php5-imagick imagemagick acl tar smbclient"
+pkg_dependencies="php5-gd php5-json php5-intl php5-mcrypt php5-curl php5-apcu php5-redis php5-ldap php5-zip php5-imagick imagemagick acl tar smbclient"
 
 #=================================================
 # COMMON HELPERS
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -3,7 +3,11 @@
 # COMMON VARIABLES
 #=================================================
 
-pkg_dependencies="php5-gd php5-json php5-intl php5-mcrypt php5-curl php5-apcu php5-redis php5-ldap php5-zip php5-imagick imagemagick acl tar smbclient"
+pkg_dependencies="php5-gd php5-json php5-intl php5-mcrypt php5-curl php5-apcu php5-redis php5-ldap php5-imagick imagemagick acl tar smbclient"
+
+if [ "$(lsb_release --codename --short)" == "stretch" ]; then
+	pkg_dependencies="$pkg_dependencies php-zip"
+fi
 
 #=================================================
 # COMMON HELPERS
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -5,7 +5,7 @@
 
 pkg_dependencies="php5-gd php5-json php5-intl php5-mcrypt php5-curl php5-apcu php5-redis php5-ldap php5-imagick imagemagick acl tar smbclient"
 
-if [ "$(lsb_release --codename --short)" == "stretch" ]; then
+if [ "$(lsb_release --codename --short)" != "jessie" ]; then
 	pkg_dependencies="$pkg_dependencies php-zip"
 fi
 


JimboJoe:
What about the discussion and proposal of [this PR](https://github.com/YunoHost/yunohost/pull/381)...?

maniackcrudelis:
Hum, that's a big piece of discussion ;)
What's the point ?

Here, the install failed anyway, even without this PR. So we have to implement an if else to handle it. Not sure that the pointed PR would change anything in that case. Or maybe I missed something.

JimboJoe:
I would say the most interesting part lies in [that comment](https://github.com/YunoHost/yunohost/pull/381#issuecomment-358322345)... :wink:

maniackcrudelis:
About the migration matter ?
I'm not sure that I get the problem here.
On jessie, there's no php-zip package. On stretch there is, I agree that during the migration the package will be missing.
But, because in case of migration no package will be removed if there any dependencies, I don't think that would affect nextcloud.
About that interesting idea of an OR in the control file. Is it a working solution ? Any sources about that ? Is apt going to recheck the dependencies in case of migration ?

JimboJoe:
I guess your opinion would be much valuable on the other PR, then (I haven't any so far... /o\ )

maniackcrudelis:
So what are we doing with this PR ?

My point of view would be that this solution isn't perfect, but it would works for a new installation on stretch.
So, I think that we can merge that to allow nextcloud to be installed, and then try to find a better way to handle that issue with "optionnal" packages.

frju365:
Btw: could you add  in your reviewer the Apps_Group ? Because you won't get enough reviewers :)

alexAubin:
Uh but with @Josue-T 's PR, can't we add something like `php5-zip | php-zip` ? (assuming php5-zip exists in jessie and is required by nextcloud) ? Not sure if that addresses the issue tho :s

maniackcrudelis:
Neither php5-zip, php-zip or php7-zip are available on jessie...

alexAubin:
So if I understand correctly, what we would like to do is to add this dependency right at the time the migration to stretch is performed ? Something like this ? (Though that doesn't seem trivial... But the recommend thing might do the trick)

maniackcrudelis:
By the way, is there a migration script yet ?
I can try to work on this issue this evening or late afternoon.

maniackcrudelis:
After a migration to stretch, nextcloud is still working but it didn't move to php7...

scripts/_common.sh
```diff
@@ -5,6 +5,10 @@
 
 pkg_dependencies="php5-gd php5-json php5-intl php5-mcrypt php5-curl php5-apcu php5-redis php5-ldap php5-imagick imagemagick acl tar smbclient"
 
+if [ "$(lsb_release --codename --short)" == "stretch" ]; then
```
Josue-T: I purpose to use `"$(lsb_release --codename --short)" != "jessie"` to be compatible with the next debian version (buster). What do you think ?

JimboJoe: I agree.
And we will be able to use `ynh_get_debian_release` as soon as a release includes this merged PR https://github.com/YunoHost/yunohost/pull/373 :wink:

alexAubin:
Bumpity bump

I don't know what was the conclusion for this ... (If I understand/remember correctly, the main point being that we need php-zip to get installed somewhere during the stretch migration and it's difficult to automate this)

One thing we can do would be to add php-zip as a dependency in yunohost, or just force the install of php-zip during the migration (possibly if we detect that nextcloud is installed). I think I would prefer going with the second solution. So something like this during the core migration : 

```python
if "nextcloud" in installed_apps:
    apt_get_install("php-zip")
```

Does that sound okay ?

maniackcrudelis:
The main problem is that we need php-zip to make nextcloud works on stretch. And we can't have it before stretch.
So, your solution would works. But, my concern is it's going to work for nextcloud but only for nextcloud.
Another solution would be to do an upgrade of nextcloud after the migration. I tried that and it works.

Also, I think that we have the same issue with kanboard, see https://github.com/YunoHost-Apps/kanboard_ynh/pull/60.

One or the other solution will work for me.

alexAubin:
Okay,

it looks like installing this dependency only take ~200 ko of disk space, so we might as well just install it during the migration with no particular condition, and that should be just fine I think :sweat_smile:

Josue-T:
Hello,
I have an idea to solve properly the issue like  that. I is to add a hook in the core for each migration. By this each app how need some change in the migration could add a hook. I know that it's a lot of work but I think It could solve many issue with many app. By this we could solve the dependence issue but also the database migration (for apps with postgresql).

alexAubin:
Granted that this would be cleaner and move evolutive ... but I would also like to avoid to spend dozens of hours or work on this, as it's gonna require some code and time to test it, also since I'd really like to be able to move forward a beta-testing to avoid ending up in June (= 1 year since Stretch release) and still havent started to properly move away from Jessie :/ ...

Can you elaborate on how you would see this working / implemented ?

Regarding the postgresql database migration, I made a quick test a few weeks ago and it seem to turn out that postgresql 9.4 and 9.6 can cohabit and everything seemed to be working right after a migration. So I'm not too worried about this, and for now my thought on this is to let applications handle the migration themselves with upgrades post-migration. (Which is therefore somewhat different from this case with the php-zip, which we need to install during the migration to avoid ending up with broken apps).

Josue-T:
> So I'm not too worried about this, and for now my thought on this is to let applications handle the migration themselves with upgrades post-migration.

So, are they already a solution for the apps to add a script witch is called while we do the migration ?

alexAubin:
While we do the migration, nope. But I was referring to the upgrade scripts, which the user can run after the migration ;). (But I'm also not too comfortable with the idea of triggering a general upgrade of all the apps automatically right after the migration :s )

We could also imagine to add a script 'jessie_to_stretch' in the scripts folder, and find a way to run it, but then we also to know if there is such a script, assuming that people don't necessarily have all their apps up to date prior to the migration ... (Then again we could run a full app upgrade before the migration, but I'd like to avoid that if possible ... Not mentioning that a lot of people install community apps without enabling the community list, therefore they can't upgrade them...)

So overall, right now it really sounds like it's just about this php-zip thing, which is easily fixable without too much work. Maybe there are similar issues, but I'd really like to find them (if they exist) before engineering a solution to cover any situation :s

Josue-T:
Ok,
So my first idea was to create a directory named for example `migration-2` for the migration number 2 (probably in  `/usr/share/yunohost/hooks/`). While the user do the migration the scripts in this directory is called. By that each app witch need a specific modification could add a script in this directory.

I know that actually we are in the alpha status for stretch and we want migrate to stretch definitively as quickly as possible. I also know that the implementation of that need some work. I just think that this idea is cleaner than to add a specific rule for nexcloud in the core. I think I could also be useful in case of a community app witch need also some modification while we do the migration. I think this Idea could also be usefull for the next migrations.

I purpose that because I had this idea some day ago and late is better than never.

alexAubin:
Yes, I do agree that this would be cleaner and useful in other context (c.f. the need for apps to be able to ship custom/specific actions scripts, that could be ran from the CLI or webadmin).

But the big picture for Stretch is that there are still many other small things to test / implement before being able to release ... and I expect such a feature (or clean solution) to take several days to implement / test, and require works both in the core and apps. All for something which, at our current understanding and knowledge, is limited to this specific php-zip thing, which otherwise can be solved with 1 or 2 of code in the current migration code... At least I'm not really enthusiast about implementing myself a whole mechanism which we are not sure that we gonna need for stretch...

So right now I'd really favor a) implementing the ad-hoc / "dirty" fix, b) if we realize that this is a recurring issue for the stretch migration, then we'll see if that's worth to implement something generic ? In any case we will have time after we get rid of the stretch situation to implement this kind of thing (again c.f. the 'app actions' thing), I think...

maniackcrudelis:
Please this PR need some love

maniackcrudelis:
So, to conclude, we keep that PR as it is, and we're going to recommend to do an upgrade of this app after migration.

If we agree with that, let's go forward.

JimboJoe:
I totally agree, let's move forward! :+1:

maniackcrudelis:
Will be merged in 3 days then.

PR #99, opened 2018-03-12
Fix upgrade old version

maniackcrudelis:
## Problem
- *Upgrade from an old version fails because the patches can't be applicated.*

## Solution
- *Move the patches and restore them at their real place only for the last upgrade. We can't expect that the patches can be applicated to all versions. And they don't have to.*

## PR Status
Could be reviewed and tested.

## Validation
---
*Minor decision*
- [x] **Upgrade previous version** : JimboJoe
- [x] **Code review** : frju365
- [x] **CI succeeded** : [Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20fix_upgrade%20(Official)/)  


scripts/upgrade
```diff
@@ -191,6 +191,12 @@ do
     # Create a temporary directory
     tmpdir=$(mktemp -d)
 
+    # Patch nexcloud files only for the last upgrade.
+    if [ "$current_major_version" = "last" ]
+    then
+        cp -a ../sources/patches_last_version ../sources/patches
```
JimboJoe: Wouldn't it be cleaner if you added that line and the corresponding comment on L179...?

maniackcrudelis: Hum, yes... Why I didn't do that ...?

JimboJoe:
Will be merged in 3 days.

PR #111, opened 2018-05-04
Add multimedia support

maniackcrudelis:
## Problem
- *Users can't handle their multimedia files, which is really to bad since Nextcloud is the only web interface to do it...*

## Solution
- *Add multimedia directories support with write access*

## PR Status
Could be reviewed and tested.

## Validation
---
*Minor decision*
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- **CI succeeded** : [Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20multimedia_support%20(Official)/)  

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -19,10 +19,12 @@ exec_occ() {
       php occ --no-interaction --no-ansi "$@")
 }
 
-# Create the external storage for the home folders and enable sharing
-create_home_external_storage() {
+# Create the external storage for the given folders and enable sharing
+create_external_storage() {
+  local datadir="$1"
+  local mount_name="$2"
   local mount_id=`exec_occ files_external:create --output=json \
-      'Home' 'local' 'null::null' -c 'datadir=/home/$user' || true`
+      "$2" 'local' 'null::null' -c "datadir=$datadir" || true`
   ! [[ $mount_id =~ ^[0-9]+$ ]] \
     && echo "Unable to create external storage" >&2 \
     || exec_occ files_external:option "$mount_id" enable_sharing true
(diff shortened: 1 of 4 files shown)


scripts/_common.sh
```diff
@@ -19,10 +19,12 @@ exec_occ() {
...
 
-# Create the external storage for the home folders and enable sharing
-create_home_external_storage() {
+# Create the external storage for the given folders and enable sharing
+create_external_storage() {
+  local datadir="$1"
+  local mount_name="$2"
   local mount_id=`exec_occ files_external:create --output=json \
-      'Home' 'local' 'null::null' -c 'datadir=/home/$user' || true`
+      "$2" 'local' 'null::null' -c "datadir=$datadir" || true`
```
Josue-T: Why is it written `|| true`
Could you add a comment to explain that.

scripts/_common.sh
```diff
@@ -315,3 +317,31 @@ ynh_handle_app_migration ()  {
...
+#=================================================
+# YUNOHOST MULTIMEDIA INTEGRATION
+#=================================================
+
+# Install or update the main directory yunohost.multimedia
+#
+# usage: ynh_multimedia_build_main_dir
+ynh_multimedia_build_main_dir () {
+        wget -nv https://github.com/YunoHost-Apps/yunohost.multimedia/archive/master.zip 2>&1
+        unzip -q master.zip
```
Josue-T: Could you add a check of a checksum ?

scripts/_common.sh
maniackcrudelis: This part of the code isn't from me, and there's no comments on the original commit.
I guess that sometime this command would fails.

scripts/_common.sh
maniackcrudelis: Yes, you're right, would be better.

I'm running package check to verify my fix, and will update this PR

JimboJoe:
Will be merged in 3 days.

PR #116, opened 2018-05-14
Fix Debian Stretch dependencies (fixes #112)

### LINKED ISSUE #112: Missing dependencies on Stretch
As post in French on https://forum.yunohost.org/t/beta-testing-phase-for-yunohost-3-0-on-debian- stretch/4798/53
Nextcloud not working on Yunohost 3.0 (on debian stretch).

JimboJoe:
## Problem
- YunoHost instances migrated to Debian Stretch from Jessie seem to lack some php dependencies

## Solution
- install missing dependencies

## PR Status
Work finished.
Could be reviewed and tested.

## Validation
---
*Minor decision*
- [x] **Code review** : Maniack C (,frju365)
- **CI succeeded** : [Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20fix_stretch_dependencies%20(Official)/)  

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -6,7 +6,7 @@
 pkg_dependencies="php5-gd php5-json php5-intl php5-mcrypt php5-curl php5-apcu php5-redis php5-ldap php5-imagick imagemagick acl tar smbclient"
 
 if [ "$(lsb_release --codename --short)" != "jessie" ]; then
-	pkg_dependencies="$pkg_dependencies php-zip"
+	pkg_dependencies="$pkg_dependencies php-zip php-apcu php-mbstring php-xml"
 fi
 
 #=================================================
@@ -139,7 +139,7 @@ ynh_handle_app_migration ()  {
   fi
 
   #=================================================
-  # CHECK IF IT HAS TO MIGRATE 
+  # CHECK IF IT HAS TO MIGRATE
   #=================================================
 
   migration_process=0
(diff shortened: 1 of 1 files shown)


frju365:
I agree totally with this PR, LGTM + Code Review... but Can't we include it in the core of Yunohost ? These are very useful modules for PHP and for the Apps which use it, and I don't think including every module in each app is a good thing. I know we had already this discussion... but finally, I think there are some dependencies we could easily include in Core to avoid problems. Moreover this module don't takes RAM/SWAP or much disk space.

maniackcrudelis:
With ynh_install_app_dependencies, apps can install only the dependencies they need, and with this helper, apps can need a same dependency without any conflict. So we have a tool to install any dependency an app can need, why should we add dependencies to the core for apps, without any certitude that those apps are going to be installed by an user?

Any app should have an install instruction that defines which php dependencies are needed.
So, really, I don't think that it's a good idea to add any php dependencies not used by the core just to give an easier way to install apps.

PS: Well, it's not really clear, but I hope that you're going to get the main idea...

frju365:
Well... I understand. I know how the helper works. :-) But still, I think some dependencies are the same for several apps. It's just my point of view.

frju365:
Will be merged in three days !

JimboJoe:
I have a shorter interpretation of time :grin:

PR #115, opened 2018-05-14
Fix automatic logout from YunoHost when logging out from Nextcloud (#106)

JimboJoe:
## Problem
- The [13.0.0 PR](https://github.com/YunoHost-Apps/nextcloud_ynh/pull/90) didn't adapt the source patch for logout feature
- Moreover, since [this PR](https://github.com/YunoHost-Apps/nextcloud_ynh/pull/99), source patches weren't applied any more on install

## Solution
- fix Nextcloud source patch
- fix patches not applied on installation (use `upgrade.last.sh` script which is used in install and upgrade)
- add an empty patches directory to work around a nasty cache issue when running package_check

## PR Status
Work finished.
Could be reviewed and tested: logging out from Nextcloud should log the user out of YunoHost and redirect to the YunoHost login screen.

## Validation
---
*Minor decision*
- **Upgrade previous version** : 
- [x] **Code review** : Maniack C
- [x] **Approval (LGTM)** : Maniack C
- [x] **Approval (LGTM)** : frju365
- **CI succeeded** : [Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20fix_logout_patch%20(Official)/)  
When the PR is mark as ready to merge, you have to wait for 3 days before really merge it.

PR #113, opened 2018-05-12
Upgrade to upstream version 13.0.2

JimboJoe:
## Problem
- *The application doesn't provide the latest NextCloud version*

## Solution
- *Upgrade to version 13.0.2*

## PR Status
Work finished.
Could be reviewed and tested.

## Validation
---
*Minor decision*
- **Upgrade previous version** : 
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- **CI succeeded** : [Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/roundcube_ynh%20enh_update_13.0.2%20(Official_fork)/)  
When the PR is mark as ready to merge, you have to wait for 3 days before really merge it.

PR #117, opened 2018-05-15
Testing

JimboJoe:
## Problem
- *Users can't handle their multimedia files, which is really to bad since Nextcloud is the only web interface to do it...*
- *The application doesn't provide the latest NextCloud version*
- *The application doesn't work on Stretch*
- *The logout functionality is broken*

## Solution
- *Add multimedia directories support with write access* https://github.com/YunoHost-Apps/nextcloud_ynh/pull/111
- *Upgrade to version 13.0.2* https://github.com/YunoHost-Apps/nextcloud_ynh/pull/113
- *Fix Debian Stretch dependencies* https://github.com/YunoHost-Apps/nextcloud_ynh/pull/116
- *Fix automatic logout from YunoHost when logging out from Nextcloud* https://github.com/YunoHost-Apps/nextcloud_ynh/pull/115
## PR Status
Could be reviewed and tested.

## Validation
---
*Minor decision*
- [x] **Code review** : Maniack C except for https://github.com/YunoHost-Apps/nextcloud_ynh/pull/117/commits/ada812a79cfe0d9fb4bed918d5659e53a60dbf2f and https://github.com/YunoHost-Apps/nextcloud_ynh/pull/117/commits/628cead60fa17dccb8eaf0f0e518e5fa70557e82
- [x] **Approval (LGTM)** : Josué
- [x] **Approval (LGTM)** : Maniack C except for https://github.com/YunoHost-Apps/nextcloud_ynh/pull/117/commits/ada812a79cfe0d9fb4bed918d5659e53a60dbf2f and https://github.com/YunoHost-Apps/nextcloud_ynh/pull/117/commits/628cead60fa17dccb8eaf0f0e518e5fa70557e82
- **CI succeeded** : [Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20testing%20(Official)/)  


JimboJoe:
I propose to wait for https://github.com/YunoHost-Apps/nextcloud_ynh/pull/116 and https://github.com/YunoHost-Apps/nextcloud_ynh/pull/115 before releasing the testing.

maniackcrudelis:
Agree

JimboJoe:
OK, every PR has been merged, so we can have a testing release.

JimboJoe:
I get systematically this in `/home/yunohost.app/nextcloud/data/nextcloud.log`:
>"Following symlinks is not allowed ('\/home\/yunohost.multimedia\/jimmy\/Share' -> '\/home\/yunohost.multimedia\/share\/' not inside '\/home\/yunohost.multimedia\/jimmy\/')"

maniackcrudelis:
Yeah, Nextcloud is capricious with symbolic links, and it seems that it isn't planned to support it...

We were discussing about that [here](https://github.com/YunoHost-Apps/yunohost.multimedia/issues/6)
I'm wondering if it would be possible (and a good idea) to add this mount to nextcloud for each users.

JimboJoe:
It seems that symlinks are forbidden in Nextcloud (see [here](https://central.owncloud.org/t/exception-while-scanning-following-symlinks-is-not-allowed/1282) or [here](https://github.com/owncloud/documentation/issues/2493)), making the whole multimedia PR pointless... :disappointed: Did it work for you...?

maniackcrudelis:
That's why I used 2 separate mount
https://github.com/YunoHost-Apps/nextcloud_ynh/pull/117/files#diff-ad962c6fd65b725b71aa46c8a715127fR248

JimboJoe:
I've seen that, but as there are symlinks between them. .. So did it work for you?

maniackcrudelis:
I just tried an upgrade to the testing release, it works (almost) as expected for me.
Just it lacks some directories that I replaced myself by symbolic links because of my own architecture. But I have 2 directories, "Multimedia" and "Shared multimedia" where I can write with nextcloud, as expected.
Still it's not perfect because it lacks also the transmission directories, which are also symbolic links, and any other symbolic links will be added by other apps. But we can do anything against that since it's a nextcloud problem and they don't want to fix it.

JimboJoe:
So, should we keep it like that?
Maybe we should mention that non-blocking error message somewhere otherwise we'll maybe get some issues/forum posts on that topic...

JimboJoe:
Shall we add https://github.com/YunoHost-Apps/nextcloud_ynh/pull/119 to this testing or wait until this one is released?

maniackcrudelis:
#119 is an important fix, I think that this testing can wait and integrate this fix.
And, yes, a small warning in the readme can be great to inform users that's not an issue, but a limitation from nextcloud.

JimboJoe:
Testing period over, could you please review it?

scripts/_common.sh
```diff
@@ -6,7 +6,7 @@
 pkg_dependencies="php5-gd php5-json php5-intl php5-mcrypt php5-curl php5-apcu php5-redis php5-ldap php5-imagick imagemagick acl tar smbclient"
 
 if [ "$(lsb_release --codename --short)" != "jessie" ]; then
-	pkg_dependencies="$pkg_dependencies php-zip"
+	pkg_dependencies="$pkg_dependencies php-zip php-apcu php-mbstring php-xml"
```
Josue-T: I have found a idea to fix the issue of "how could we define that we need only this package on jessie" : 

Just write by example (for php-zip only on jessie) : `"php-zip | base-files>=9.0"`
"base-file" is : "Debian base system miscellaneous files"

I know that it's really late but I prefer to share this idea because I think it could solve many actual problem.

maniackcrudelis: Actually, I rather prefer a ` if not jessie then do that` instead of the syntax you suggest. Because, I think that is clearer to have the current syntax. And you know that I like to keep the code clear and readable ;)

maniackcrudelis:
Code review OK, except for https://github.com/YunoHost-Apps/nextcloud_ynh/pull/117/commits/ada812a79cfe0d9fb4bed918d5659e53a60dbf2f and https://github.com/YunoHost-Apps/nextcloud_ynh/pull/117/commits/628cead60fa17dccb8eaf0f0e518e5fa70557e82 which are my own commits

JimboJoe:
Each commit has already been reviewed in its contributing PR, so in my understanding the code review in this PR only consists to check possible interactions between the different contributing PRs...

scripts/_common.sh
Josue-T: Yes I really understand your point of view but with that we don't need to upgrade the app after the migration to stretch. By that we might probably avoid some migration problem about from the dependence. It's from my point of view the main advantage of that.

But it's stay a suggestion.

maniackcrudelis: Do you mean that a simple apt upgrade will install the missing packages during the migration to stretch ?

Josue-T: Yes probably. It's the reason because I purposed that.

maniackcrudelis: "probably" ? ;)
Could be nice to be sure about that.

Josue-T: I say probably because I did't try the migration but I have tried [in friendica package](https://github.com/YunoHost-Apps/friendica_ynh/commit/2706331022719e7ed355baa85fd38a4ac9ba5e6c) the install in jessie and stretch and it was good.

About the migration yes I haven't tried yet but I really don't see any reason that I won't work because with this concept we Install always the same package and it's apt which will analyse the dependence.

Josue-T: Anyway If you are really interested by that I could do a migration test but I don't want to lose my time for an unused things.

maniackcrudelis: We know that this issue could be a real problem during the migration, because in fact, it needs to upgrade each apps to install the dependencies.
So, if your solution works, it could be useful.

Josue-T: Ok, is it ok for you if I do a migration test tomorrow with friendica which implement already that concept to validate completely that idea ?

maniackcrudelis: Yes.
However, would be even better to try with nextcloud, because we know that currently it doesn't work in case of migration without doing an upgrade.
But, nevertheless, the main thing is to be sure that the missing packages will be installed during the migration.

alexAubin:
Are we good to go then ? Time is running and a stable for stretch is to be expected Soon™ :s

scripts/_common.sh
Josue-T: I tried a migration and the problem that I had now was that I didn't updated all "php5" dependence. I think to solve completely the problem we need to use something like that for each "php5" dependences : 
`php5-curl|php-curl` to say to apt to use the php7 version in stretch for each dependence.
I could try a new time a migration I your stay interested on that.

With something like that it should work : 
```
pkg_dependencies="php5-gd|php-gd php5-json|php-json php5-intl|php-intl php5-mcrypt|php-mcrypt php5-curl|php-curl php5-apcu|php-apcu php5-redis|php-redis php5-ldap|php-ldap php5-imagick|php-imagick imagemagick acl tar smbclient php-zip|base-files(<<9.0) php-apcu|base-files(<<9.0) php-mbstring|base-files(<<9.0) php-xml|base-files(<<9.0)"
```

maniackcrudelis: Why not since nextcloud is official.
But, we shouldn't expect anyone else to do the same, this syntax is far more too complex for a simple list of dependencies...
If it works though, a comment would much appreciate.

Josue-T: So, I did a new test and know it work better. The last problem is that apt is lazy so it won't install by example `php7.0-gd` as long as `php5-gd` is installed. But for the others it's ok.

JimboJoe: So when migrating, php5-* packages are kept and this package just can't work with dependencies handled like this, can it?

Josue-T: I tested with roundcube and nextcloud and after the migration both app worked again without any upgrade but I think we cant guarantee that all package are installed. The solution that I found to force apt to install all new dependence is to hold the package (nextcloud_ynh_deps) and remove the php5-* package and in this case apt will automatically install the new dependence.

maniackcrudelis: Not sure that would be better than an simple upgrade of the app in this case...

Josue-T: > Not sure that would be better than an simple upgrade of the app in this case...

Me too

maniackcrudelis: Considering that it's only a problem with the migration to stretch, I think really that we can keep our current if/else. We're not going to have a migration everyday. So let's keep it clear and easy.

JimboJoe:
OK, there seems to be consensus on the fact that the dependencies upgrade is OK like this.
As the PR is validated, let's release!

Issue #109, opened 2018-04-28
Check if there is enough space in /tmp before attempting upgrade


roukydesbois:
Hello,
I am running Yunohost on a Olimex board. By default when you install Armbian, a separate partition of 500M is created for /tmp. So, when I tried to upgrade Nextcloud, the upgrade failed because there was not enough space on /tmp to download and/or extract the archive containing Nextcloud. The nextcloud database was then removed because the upgrade failed.
So I think it would be useful to add a check before the upgrade to see if there is enough space on /tmp, halting the upgrade if it is not big enough.

JimboJoe:
Fixed by https://github.com/YunoHost-Apps/nextcloud_ynh/pull/127

PR #127, opened 2018-06-24
Don't use /tmp anymore

maniackcrudelis:
## Problem
- *There's often problems with /tmp too small and blocking the upgrade of Nextcloud*
  - https://forum.yunohost.org/t/official-app-nextcloud/4104/70
  - https://forum.yunohost.org/t/tmp-too-small-to-update-nextcloud-app/5069

## Solution
- *Use a temporary directory in /var/www instead of /tmp since when /tmp is too small is generally because of an old YunoHost iso.*
- *Fix https://github.com/YunoHost-Apps/nextcloud_ynh/issues/109*

## PR Status
- [x] Code finished.

## Validation
---
*Minor decision*
- **CI succeeded** : 
[Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20donotusetmp%20(Official)/)  

### DIFF
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -376,6 +376,8 @@ ynh_smart_mktemp () {
                 local tmpdir=/   
         elif is_there_enough_space /home; then
                 local tmpdir=/home
+        else
+		ynh_die "Insufficient free space to continue..."
         fi
 
         echo "$(sudo mktemp --directory --tmpdir="$tmpdir")"
--- a/scripts/_common.sh
+++ b/scripts/_common.sh
@@ -356,3 +356,27 @@ ynh_multimedia_addaccess () {
         groupadd -f multimedia
         usermod -a -G multimedia $user_name
 }
+
+ynh_smart_mktemp () {
+        local min_size="${1:-300}"
+        # Transform the minimum size from megabytes to kilobytes
+        min_size=$(( $min_size * 1024 ))
+
+        # Check if there's enough free space in a directory
+        is_there_enough_space () {
+                local free_space=$(df --output=avail "$1" | sed 1d)
+                test $free_space -ge $min_size
+        }
+
+        if is_there_enough_space /tmp; then
+                local tmpdir=/tmp
+        elif is_there_enough_space /var; then
+                local tmpdir=/var
+        elif is_there_enough_space /; then
+                local tmpdir=/   
+        elif is_there_enough_space /home; then
+                local tmpdir=/home
+        fi
+
+        echo "$(sudo mktemp --directory --tmpdir="$tmpdir")"
+}
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -189,7 +189,7 @@ do
     ynh_replace_string "__SHA256_SUM__" "$nextcloud_source_sha256" "../conf/app.src"
 
     # Create a temporary directory
-    tmpdir=$(mktemp -d)
+    tmpdir="${final_path}_temp_upgrade_dir"
 
     # Install the next nextcloud version in $tmpdir
     ynh_setup_source "$tmpdir"
@@ -208,6 +208,7 @@ do
     # Replace the old nextcloud by the new one
     ynh_secure_remove "$final_path"
     mv "$tmpdir" "$final_path"
+    ynh_secure_remove "$tmpdir"
 
     # Set write access for the following commands
     chown -R $app: "$final_path" "$datadir"
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -189,7 +189,7 @@ do
     ynh_replace_string "__SHA256_SUM__" "$nextcloud_source_sha256" "../conf/app.src"
 
     # Create a temporary directory
-    tmpdir="${final_path}_temp_upgrade_dir"
+    tmpdir="$(ynh_smart_mktemp 300)"
 
     # Install the next nextcloud version in $tmpdir
     ynh_setup_source "$tmpdir"


JimboJoe:
Great to handle that case!
While the PR is very legit as it is, I wonder if it would be better to have ynh_mktemp and ynh_rmtemp helpers: here we assume there is more free space on /var than /tmp, but maybe that's not always the case? Why not look for free space on /tmp, and if less than 300Mb for example, use /var (and even checking free space in that fallback case)?
I'm asking that because we'll need to handle this in every app.

maniackcrudelis:
Yes, could be an interesting idea indeed.
Do you want to work on this matter ?

JimboJoe:
I unfortunately just don't have much time at the moment... :cry:
Would just deciding on the helper names and moving your exact code in them be a good start for now...?

maniackcrudelis:
The CI fails because of this https://github.com/YunoHost-Apps/nextcloud_ynh/pull/129.
Can be ignored.

JimboJoe:
When looking at [mktemp man page](https://www.gnu.org/software/autogen/mktemp.html), I've seen that we can specify the parent directory, and that it is even using the `$TMPDIR` environment variable (just tested it)...
Wouldn't coding a workaround in the core to launch the scripts with a different `$TMPDIR` when `/tmp` is short be a more general solution instead of changing every app? :thinking: 
cc @Psycojoker @alexAubin

maniackcrudelis:
The issue is not really about where the scripts are executed, but more about where nextcloud install the source during the upgrade.
And it's quite specific to nextcloud in that case.

JimboJoe:
> The issue is not really about where the scripts are executed, but more about where nextcloud install the source during the upgrade.

Not sure to understand... `$TMPDIR` defines the directory where `mktemp` creates files/directories, which is the very precise thing we want to change, don't we?

>And it's quite specific to nextcloud in that case.

How is this change specific to nextcloud? Won't we face that problem with any package that downloads a big file?

maniackcrudelis:
> How is this change specific to nextcloud? Won't we face that problem with any package that downloads a big file?

Not really, we have a specific problem with nextcloud because of this line https://github.com/YunoHost-Apps/nextcloud_ynh/blob/donotusetmp/scripts/upgrade#L194-L195
A general solution (which is already deployed I think) is to not use that crappy partitioning with a separate (and too small) /tmp.

JimboJoe:
Unfortunately there are other apps in this case, and even official ones... https://github.com/YunoHost-Apps/piwigo_ynh/blob/master/scripts/install#L88-L89

maniackcrudelis:
In that case, considering we may encounter that situation, since there was some YunoHost installed with this default partioning. We should either use a helper to handle that case, or propose a fix to move /tmp to the main partition. Which seems to me as a better idea.

JimboJoe:
I drew the same conclusion right in this same time! :wink: 
I wonder if fixing that /tmp partition (if needed) shouldn't be achieved in a core migration (wouldn't a simple symbolic link /tmp --> /home/tmp do the trick in that case...?)
cc @Psycojoker @alexAubin

maniackcrudelis:
Wouldn't it be easier to just remove the mount in fstab for /tmp.
Since the directory already exist.
Maybe there's some permissions to fix as well, but I don't think there's anything else to do.

The only thing will be that without a tool like gparted, the partition won't be used anymore. But it's only about 300Mo.

JimboJoe:
Provided / partition is big enough, that solution looks fairly good!

maniackcrudelis:
About this PR, what should we do ?
I don't think tweaking the fstab automatically would be a great idea.
Still, there's a potential issue.
Maybe we could create a quick (and dirty ?) helper to handle that case, use then another directory to do the upgrade. And print a warning to explain what happen and how to fix it.

Josue-T:
> I don't think tweaking the fstab automatically would be a great idea.

+1 . There are too many different architecture that we can't predict.

And how about using `/var/tmp` for `$TMPDIR` in the core ?

In any case I'm also for the idea to fix this in the core and not in the app.

maniackcrudelis:
/var is also a problem if the default debian partitions are used.
Using another directory with `--tmpdir` could be a less invasive solution for a helper. It would just prevent an error, and let the admin be informed that there were a problem and how to fix it.

maniackcrudelis:
I made a "quick and dirty" helper to find a place with enough free space to create a temporary directory.

maniackcrudelis:
I discussed about that issue with Aleks, who told me to open a issue. So I did.
But, I made this helper with the idea that it could be helpful even after the /tmp problem will be solved in the core.

maniackcrudelis:
Can we have review on this PR please.
@YunoHost-Apps/apps-group

anmol26s:
We can have two folders nextcloud and nextcloud-version with nextcloud having the unchanged config file and folders in it. The nextcloud-version could have symbolic link to all files inside the folder to reflect in nextcloud folder. This way we only need to remove nextcloud-version and replace with new version.

Further I have seen generally when the upgrade fails(because of say dependency) the restore script fails too(has happened in past with nextcloud) and the end user has no solution but to edit the restore script inside the backup folder to make the app working again with his previous data.
 In future the remove script can have block which will not run if the upgrade script fails(soft remove). Whenever the remove script would be running from cmd or  admin web-ui it would run this block(hard remove). This way on upgrade and restore fail the fresh install will make the app working again with previous data. 

But its just a proposal, we still need to check the space where the tmp folder should be created, as this is used by many apps for now.

anmol26s:
> Could be a solution, to implement though... But it could be a little more difficult with the "apps" directory https://github.com/YunoHost-Apps/nextcloud_ynh/blob/master/scripts/upgrade#L206

I agree folders like apps is little tricky to handle.

> If you know why the restore script fails, it would be more interesting to fix it.
> Not sure to understand, by the way, why nextcloud can't works with the previous data.

I was talking in general for all the apps. Sometimes upgrade and restore fails for the same reason(a migration or build specific error) and the remove script removes the app with restore failing too. User has no choice but to fresh install the app in that case,if he don't know how to fix the restore script in backup.


It would be good to have an error message if all the folder /tmp , /var,/,/home don't have space to have temp folder in yunohost die.

maniackcrudelis:
up

maniackcrudelis:
I already tested it. But, any further tests are welcome if you want.

maniackcrudelis:
Bump

JimboJoe:
CI test broken, and @frju365 approval only in the PR text...

maniackcrudelis:
This PR is too old for the CI...
The test has been removed...

Would you like a new one ?

JimboJoe:
No, I was just trying to enforce the process :laughing: 
I'm OK to merge this in 3 days.

PR #160, opened 2019-02-07
Upgrade to upstream version 15.0.4

JimboJoe:
## Problem
- *The app isn't up-to-date*
- *Due to recent [core changes](https://github.com/YunoHost/yunohost/commit/8cb029a55e471e1ece3a8a2d7bba00975a6f2d17), security headers are declared twice, raising warnings in Nextcloud administration panel*

## Solution
- *Upgrade to latest version 15.0.4*
- *Adapt nginx configuration to the specific YunoHost context*

## PR Status
- [ ] Code finished.

## Validation
---
*Minor decision*
- [ ] **Code review** : 
- [x] **Approval (LGTM)** : Anmol
- **CI succeeded** : 
[Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20enh_update_15.0.4%20(Official)/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Nextcloud for YunoHost
 own data. A personal cloud which run on your own server. With Nextcloud
 you can synchronize your files over your devices.
 
-**Shipped version:** 15.0.2
+**Shipped version:** 15.0.4
 
 [![Install Nextcloud with YunoHost](https://install-app.yunohost.org/install-with-yunohost.png)](https://install-app.yunohost.org/?app=nextcloud)
 ![](https://raw.githubusercontent.com/nextcloud/screenshots/master/files/Files%20Overview.png)
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -13,23 +13,23 @@ location ^~ __PATH__ {
   }
 
   # Add headers to serve security related headers
-  add_header Strict-Transport-Security "max-age=15768000;";
-  add_header X-Content-Type-Options nosniff;
-  add_header X-XSS-Protection "1; mode=block";
-  add_header X-Robots-Tag none;
-  add_header X-Download-Options noopen;
-  add_header X-Permitted-Cross-Domain-Policies none;
-  add_header Referrer-Policy no-referrer;
+  more_set_headers "Strict-Transport-Security: max-age=15768000";
+  more_set_headers "X-Content-Type-Options: nosniff";
+  more_set_headers "X-XSS-Protection: 1; mode=block";
+  more_set_headers "X-Robots-Tag: none";
+  more_set_headers "X-Download-Options: noopen";
+  more_set_headers "X-Permitted-Cross-Domain-Policies: none";
+  more_set_headers "Referrer-Policy: no-referrer";
 
   # Set max upload size
   client_max_body_size 10G;
   fastcgi_buffers 64 4K;
-  
+
   # Extend timeouts
   client_body_timeout 60m;
   proxy_read_timeout 60m;
   fastcgi_read_timeout 60m;
-  
+
   # Disable gzip to avoid the removal of the ETag header
   gzip off;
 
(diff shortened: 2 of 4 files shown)


JimboJoe:
I recommend to wait for this PR before releasing the stable version.

anmol26s:
What do ` more_set_headers`  do?

JimboJoe:
Please see [here](https://github.com/openresty/headers-more-nginx-module#more_set_headers) for `more_set_headers` documentation. Basically, this change ensures `more_set_headers` is used in the core and in the Nextcloud app, so that `more_set_headers` can deal with the duplicate header declaration to kind of "merge" them so we have only one... which avoids the Nextcloud warning (which can somewhat be seen as a bug).

alexAubin:
Ah, I forgot about that thing, but what do you guys think about adding a check in package_linter to warn packagers about the use of add_header and recommend more_set_headers instead ?

(By the way we might also be able to add a check about the path traversal stuff I believe)

anmol26s:
This would be great to have warning in package_linter. because I have few packages (I don't  remember) in which add_header to more_set_headers.

PS: Is there a way to remove/ignore header in the nginx configuration from the app nginx configuration?

JimboJoe:
Well, that could prove useful but:
* I'm still uncomfortable/disappointed that we need to change every app
* This change seems only to work around a Nextcloud bug, as double declaring some security headers doesn't look harmful to me (and to securityheaders.com)

So, just to be on the safe side, are we sure generalizing this change is the way to go... ? :thinking:

alexAubin:
Eh well I don't know exactly :/ My understanding is that `add_header`, if used in a "child" location, will cause this location to ignore *all* headers from the parent ... This was identenfied a few months ago. This can has some security implications because some stuff like HSTS will be effective for the root-level but not for the subpaths / applications.

Anyway, we indeed recently `add_header` to `more_set_headers` in the core and my understanding is that the headers from the parents are therefore not supressed anymore by the child locations ? But that's *just a guess* because one would usually do the other way around (use `add_header` in parents and `more_set_header` in child). But in any case, just using `more_set_header` everywhere should be way more consistent ...

So to summarize my understanding : 
- the new `more_set_header` from the core are now indeed propagated to the child, that's the expected behavior
- this might trigger issues with many apps for which the headers config is important :/ 
- fixes need to be done on a per-app basis, because we can't really change nginx's syntax behavior ... :/

(But again I don't have a clear understanding of all this so :sweat_smile: )

alexAubin:
@anmol26s 

>PS: Is there a way to remove/ignore header in the nginx configuration from the app nginx configuration?

I am not sure, but maybe `more_clear_header` (doc : https://github.com/openresty/headers-more-nginx-module#more_clear_headers ) is what you're looking for ... to be checked.

JimboJoe:
Thanks for your detailed answer! :heart: 
Given your explanations and the link you provided with package_linter PR, I'm OK that what we're doing seems the best to do on the short term. Maybe nginx will take care of this concern later on (??).
I propose to merge this PR in 3 days.

alexAubin:
Btw if some people didn't read about the issue behind `add_header` in child locations, have a look at : https://www.peterbe.com/plog/be-very-careful-with-your-add_header-in-nginx

I couldn't find another article explaining a concrete example that you might end up with a website thinking you have security stuff (e.g. HSTS) enabled, but because you used `add_header` in your sublocations, in fact you don't ...

>Given your explanations and the link you provided with package_linter PR, I'm OK that what we're doing seems the best to do on the short term. Maybe nginx will take care of this concern later on (??).

I doubt :/ The more I learn about that kind of stuff, the more it seems like nginx like to do weird counter-intuitive stuff ... That the case with `add_header`, with the path-traversal issue (which we might also want to report in the package linter btw), and with gzip_types that someone told me about tonight. There are probably some others ... I don't think they'll ever change those behavior as it would create backward incompatibilities :/ ...

alexAubin:
Bumpz, dunnoy what's the status of this ... ?

To me this looks like the good way to do things ?

PR #163, opened 2019-02-16
Package upgrade

maniackcrudelis:
## Problem
- *Global upgrade of the package.*

## PR Status
- [x] Code finished.
- [x] Tested with Package_check.
- [x] Upgrade from last version tested.

## Validation
---
*Minor decision*
- **CI succeeded** : 
[Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20package_upgrade%20(Official)/)  

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -6,8 +6,11 @@ location = /.well-known/caldav {
 }
 
 location ^~ __PATH__ {
+
+  # Path to source
   alias __FINALPATH__/;
 
+  # Force usage of https
   if ($scheme = http) {
     rewrite ^ https://$server_name$request_uri? permanent;
   }
@@ -24,12 +27,12 @@ location ^~ __PATH__ {
   # Set max upload size
   client_max_body_size 10G;
   fastcgi_buffers 64 4K;
-  
+
   # Extend timeouts
   client_body_timeout 60m;
   proxy_read_timeout 60m;
   fastcgi_read_timeout 60m;
-  
+
   # Disable gzip to avoid the removal of the ETag header
   gzip off;
 
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -70,7 +70,7 @@ location ^~ __PATH__ {
     fastcgi_param HTTPS on;
     fastcgi_param modHeadersAvailable true;
     fastcgi_param REMOTE_USER $remote_user;
-    fastcgi_pass unix:/var/run/php5-fpm-__NAME__.sock;
+    fastcgi_pass unix:/var/run/php/php7.0-fpm-__NAME__.sock;
     fastcgi_intercept_errors on;
   }
 
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -16,13 +16,13 @@ location ^~ __PATH__ {
   }
 
   # Add headers to serve security related headers
-  add_header Strict-Transport-Security "max-age=15768000;";
-  add_header X-Content-Type-Options nosniff;
-  add_header X-XSS-Protection "1; mode=block";
-  add_header X-Robots-Tag none;
-  add_header X-Download-Options noopen;
-  add_header X-Permitted-Cross-Domain-Policies none;
-  add_header Referrer-Policy no-referrer;
+  more_set_headers "Strict-Transport-Security: max-age=15768000";
+  more_set_headers "X-Content-Type-Options: nosniff";
+  more_set_headers "X-XSS-Protection: 1; mode=block";
+  more_set_headers "X-Robots-Tag: none";
+  more_set_headers "X-Download-Options: noopen";
+  more_set_headers "X-Permitted-Cross-Domain-Policies: none";
+  more_set_headers "Referrer-Policy: no-referrer";
 
   # Set max upload size
   client_max_body_size 10G;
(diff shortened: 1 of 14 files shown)


scripts/backup
```diff
@@ -77,11 +70,7 @@ ynh_backup "/etc/cron.d/$app"
...
-if [ -z $backup_core_only ]
-then
-	ynh_backup "/home/yunohost.app/${app}/data"
-else
-	echo "Data dir will not be saved, because backup_core_only is set." >&2
-fi
+# The 1 parameter indicates the directory is "big",
+# so that it won't be backed up before upgrade
+# This argument has to be the third one.
+ynh_backup "/home/yunohost.app/${app}/data" "/home/yunohost.app/${app}/data" 1
```
maniackcrudelis: This part should be really tested, as well for Jirafeau.
I wonder what will going on with a restoration for which a directory wasn't backup.
If the directory is simply missing, it could be a real problem.

JimboJoe: Hm, how is it different from the situation without this helper?
We only omit backing up this directory when upgrading, and the restore script is adapted to that. How can that fail...?

conf/nginx.conf
```diff
@@ -67,7 +70,7 @@ location ^~ __PATH__ {
     fastcgi_param HTTPS on;
     fastcgi_param modHeadersAvailable true;
     fastcgi_param REMOTE_USER $remote_user;
-    fastcgi_pass unix:/var/run/php5-fpm-__NAME__.sock;
+    fastcgi_pass unix:/var/run/php/php7.0-fpm-__NAME__.sock;
```
JimboJoe: In preparation to Debian Buster (and PHP 7.3),  we could maybe remove the ".0"...?

scripts/_common.sh
```diff
@@ -330,16 +313,20 @@ ynh_handle_app_migration ()  {
...
-        local ynh_media_release="v1.0"
-        local checksum="4852c8607db820ad51f348da0dcf0c88"
+        local ynh_media_release="v1.2"
+        local checksum="806a827ba1902d6911095602a9221181"
 
         # Download yunohost.multimedia scripts
         wget -nv https://github.com/YunoHost-Apps/yunohost.multimedia/archive/${ynh_media_release}.tar.gz 
 
-        # Verify checksum
+        # Check the control sum
```
JimboJoe: Wasn't "checksum" more appropriate...? :thinking:

scripts/install
```diff
@@ -123,50 +124,62 @@ ynh_add_fpm_config
...
 #=================================================
 # CONFIGURE NEXTCLOUD
 #=================================================
+ynh_print_info "Configuring nextcloud..."
 
 # Ensure that UpdateNotification app is disabled
 exec_occ app:disable updatenotification
 
-# Enable plugins
+# Enable ldap plugins
```
JimboJoe:
```suggestion
# Enable ldap plugin
```

scripts/_common.sh
maniackcrudelis: Yep maybe, but that would be fixed in the helper itself. That's just a copy of it.

conf/nginx.conf
maniackcrudelis: Yep, that's a good idea.
Should be done in the example app first though.

scripts/backup
maniackcrudelis: That's indeed in that restore script.
Maybe I was thinking especially about Jirafeau, don't remember...

maniackcrudelis: Confirm that it doesn't work for Jirafeau, and not_mandatory isn't yet available for restore helper

manifest.json
```diff
@@ -6,20 +6,20 @@
...
+    "version": "15.0.4~ynh2",
     "url": "https://nextcloud.com",
     "license": "AGPL-3.0",
     "maintainer": {
         "name": "YunoHost Contributors",
         "email": "apps@yunohost.org"
     },
     "requirements": {
-        "yunohost": ">= 2.7.2"
+        "yunohost": ">= 3.0"
```
JimboJoe:
```suggestion
        "yunohost": ">= 3.2.0"
```

scripts/install
```diff
@@ -181,13 +194,26 @@ exec_occ ldap:test-config \'\' \
 # MOUNT HOME FOLDERS AS EXTERNAL STORAGE
 #=================================================
 
+# Define a function to add an external storage
+# Create the external storage for the given folders and enable sharing
+create_external_storage() {
```
JimboJoe: I'm still a bit puzzled about the balance between:
* this duplicated function (deleted from `_common.sh`), which is quite atomic and well defined as such
* experimental helpers, which have the "right" to be factorized and hence defined in `_common.sh`...

The only important difference I see is that this function is specific to this application (Nextcloud), whereas experimental helpers are designed to be versatile.

Is that the general rule in your mind, @maniackcrudelis ?

I thought about this when planning to apply the latest template to discourse package, and stumbled upon these [internal helpers](https://github.com/YunoHost-Apps/discourse_ynh/blob/master/scripts/_common.sh#L30-L69). I so understand that I need to publish them as experimental helpers to be able to factorize my code...? :wink:

Cc @YunoHost-Apps/apps-group

maniackcrudelis: That's indeed the idea, to factorize only parts of the code that's will be reused by every packagers. So we won't have to read the code to know what it's going on in it.

In opposition with specific functions for which you don't know what's into it.
Considering also that Experimental helpers are supposed to be official by the time.

Here I indeed defined a function into the install script to not rewrite the same thing many times. But the function is where we're going to to use it. Not elsewhere, so the code stays easily readable.

JimboJoe:
:up:

JimboJoe:
Bump again @YunoHost-Apps/apps-group, could be useful to base further PR on the refactorized code... :wink:

maniackcrudelis:
Will be merged in 3 days.

PR #166, opened 2019-02-24
Quote around variables to avoid miserably crashing if variable is empty

alexAubin:
## Problem

c.f. https://forum.yunohost.org/t/nextcloud-mise-a-jour-impossible/7199/25 ... Turns out variable was empty

## Solution

Having quotes around variables where you can't 100% sure won't be empty doesn't cost much :s 

## PR Status
- [X] Code finished.

## Validation
---
*Minor decision*
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- **CI succeeded** : 

[Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20PR166%20(Official_fork)/)

### DIFF
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -49,7 +49,7 @@ fi
 current_version=$(grep OC_VersionString "$final_path/version.php" | cut -d\' -f2)
 current_major_version=${current_version%%.*}
 
-if [ $current_major_version -gt 11 ]
+if [ "$current_major_version" -gt 11 ]
 then
     # Inform the backup/restore process that it should not save the data directory
     ynh_app_setting_set $app backup_core_only 1
@@ -73,7 +73,7 @@ ynh_abort_if_errors
 #=================================================
 
 ynh_handle_app_migration "owncloud" "owncloud_migration"
-if [ $migration_process -eq 1 ]
+if [ "$migration_process" -eq 1 ]
 then
     # If a migration has been perform
     # Reload some values changed by the migration process
@@ -270,7 +270,7 @@ exec_occ config:system:get logout_url >/dev/null 2>&1 \
 #=================================================
 
 # Enable External Storage and create local mount to home folder as needed
-if [ $user_home -eq 1 ]; then
+if [ "$user_home" -eq 1 ]; then
     exec_occ app:enable files_external
     exec_occ files_external:list --output=json \
 	| grep -q '"storage":"\\\\OC\\\\Files\\\\Storage\\\\Local"' \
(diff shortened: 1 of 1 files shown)


alexAubin:
(Not sure what to do with the build thing @.@)

JimboJoe:
I'm not a bash expert, but doesn't it implicitly convert to string, whereas we're comparing to an integer...?

alexAubin:
Good question, but apparently running : 

```
[ "3" -eq 3 ] && echo "yay!" || echo "uhoh"
```

displays "yay" :wink: It's not 100% clear to me but it seems to me that there's no clear notion of string / int in bash. Just that `-eq` performs a *numerical* comparison instead of alphabetical one ?

Josue-T:
Maybe there are a little difference, see:

```
$ [[ 2 -eq 1+1 ]] && echo hello
hello
$ [[ 2 == 1+1 ]] && echo hello
$ 
```

alexAubin:
Yes, `==` is string comparison. Though surprisingly, `-eq` (arithmetic comparison) is able to perform the `1+1` sum :o.

Anyway, for that we're talking about, notice that : 

```
[ "3" == 3 ] && echo "yay!"
```

also prints yay!

maniackcrudelis:
First, and AGAIN, do not compare [[ ]] and [ ]. That's 2 different things !

-eq is indeed a numeric check, but there's no type in bash. So it doesn't matter if it's a string or not. But it matters if it's not only digit.

>  Attention : ./upgrade: ligne 52 : [: -gt : opérateur unaire attendu

That doesn't mean that the variable was empty, just that it wasn't only digit. We would have a complete log to know what contains the variable.
By the way, did you check that with double quote it works ? I'm not sure.

maniackcrudelis:
As expected, with or without double quotes, if fails with an empty variable or a non digit variable.

alexAubin:
Indeed :sweat_smile: : 

```
$ [ "" -eq 3 ]
bash: [: : integer expression expected
```

The origin of the issue was : 

>Attention : grep: /var/www/nextcloud/version.php: Aucun fichier ou dossier de ce type
>Attention : ./upgrade: ligne 52 : [: -gt : opérateur unaire attendu

c.f. here : https://github.com/YunoHost-Apps/nextcloud_ynh/blob/master/scripts/upgrade#L49

So I guess in fact a proper fix would be to default to some other value for `$current_major_version` ? (Apart from the fact that not having a `version.php` is probably quite weird in the first place..)

maniackcrudelis:
> Apart from the fact that not having a `version.php` is probably quite weird in the first place..

I'm more concerned about that. The problem isn't that the test has failed on an empty variable, but that the file was missing.

maniackcrudelis:
Close, as it doesn't solve the issue. And doesn't work actually.

PR #174, opened 2019-03-07
Speedup chown and chmod

Mickael-Martin:
## Problem
The upgrade is very slow for installation with many files (1hour for me if I have not out of memory)

## Solution
Change permissions/owner only if the file have bad permission (read attributes is faster than read and write all files).
Please try with testing instance this PR, this patch make the upgrade for me in 20 minutes.

## PR Status
- [ ] Code finished.

## Validation
---
*Minor decision*
- [x] **Code review** : Kay0u
- **CI succeeded** : 
[Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20PR172%20(Official_fork)/)  

### DIFF
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -160,7 +160,7 @@ datadir="/home/yunohost.app/${app}/data"
 # Set write access for the following commands
 chown -R $app: "$final_path" 
 # Change of owner and group if files in datadir does not have the right permissions
-find "${datadir}/" \! -user $app -o \! -group $app -exec chown $app: {} \;
+find "$datadir/" \! -user $app -o \! -group $app | xargs -d '\n' chown $app:
 
 # Print the current version number of nextcloud
 exec_occ -V
@@ -217,7 +217,7 @@ do
     # Set write access for the following commands
     chown -R $app: "$final_path" 
     # Change of owner and group if files in datadir does not have the right permissions
-    find "${datadir}/" \! -user $app -o \! -group $app -exec chown $app: {} \;
+    find "$datadir/" \! -user $app -o \! -group $app | xargs -d '\n' chown $app:
 
     # Upgrade Nextcloud (SUCCESS = 0, UP_TO_DATE = 3)
     exec_occ maintenance:mode --off
@@ -335,13 +335,13 @@ ynh_multimedia_addaccess $app
 # Fix app ownerships & permissions
 chown -R $app: "$final_path" 
 # Change of owner and group if files in datadir does not have the right permissions
-find "${datadir}/" \! -user $app -o \! -group $app -exec chown $app: {} \;
-find "${final_path}/" -type f -print0 | xargs -0 chmod 0644
-find "${final_path}/" -type d -print0 | xargs -0 chmod 0755
+find "$datadir/" \! -user $app -o \! -group $app | xargs -d '\n' chown $app:
+find "$final_path/" -type f | xargs -d '\n' chmod 0644
+find "$final_path/" -type d | xargs -d '\n' chmod 0755
 # Change permissions if the file in datadir does not have the right permissions
-find "${datadir}/" -type f \! -perm 0640 -print0 | xargs -0 chmod 0640
+find "$datadir/" -type f \! -perm 0640 | xargs -d '\n' chmod 0640
 # Change permissions if directories in datadir does not have the right permissions
-find "${datadir}/" -type d \! -perm 0750 -print0 | xargs -0 chmod 0750
+find "$datadir/" -type d \! -perm 0750 | xargs -d '\n' chmod 0750
 chmod 640 "${final_path}/config/config.php"
 chmod 755 /home/yunohost.app
 
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -158,7 +158,8 @@ last_version=$next_version
 datadir="/home/yunohost.app/${app}/data"
 
 # Set write access for the following commands
-chown -R $app: "$final_path" "$datadir"
+chown -R $app: "$final_path" 
+find ${datadir}/ \! -user $app -o \! -group $app -exec chown $app: {} \;
 
 # Print the current version number of nextcloud
 exec_occ -V
@@ -213,7 +214,8 @@ do
     ynh_secure_remove "$tmpdir"
 
     # Set write access for the following commands
-    chown -R $app: "$final_path" "$datadir"
+    chown -R $app: "$final_path" 
+    find ${datadir}/ \! -user $app -o \! -group $app -exec chown $app: {} \;
 
     # Upgrade Nextcloud (SUCCESS = 0, UP_TO_DATE = 3)
     exec_occ maintenance:mode --off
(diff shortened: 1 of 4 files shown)


scripts/upgrade
```diff
@@ -158,7 +158,8 @@ last_version=$next_version
 datadir="/home/yunohost.app/${app}/data"
 
 # Set write access for the following commands
-chown -R $app: "$final_path" "$datadir"
+chown -R $app: "$final_path" 
+find ${datadir}/ \! -user $app -o \! -group $app -exec chown $app: {} \;
```
maniackcrudelis: It could use some explanation.

Mickael-Martin: datadir can have many files, make chown on any take time because it's a write operation on any file.
Checking if file don't have the right owner can reduce this time.

maniackcrudelis: "It" not "I"
I don't need any explanations about bash. But this code should be commented.

Mickael-Martin: Done ! (sorry for my english...)

maniackcrudelis:
Did you try this code of directories with space in names ?

Mickael-Martin:
> Did you try this code on directories with space in names ?
Fatal omission... I fix it now.

Mickael-Martin:
I have added double quotes on final_path too, it's ok ?

maniackcrudelis:
You tried it, so is it ok now ?

During an install, you don't need that since the directory is empty.

Mickael-Martin:
I have tested only this part with the fix with double quote.
I have used this code for my personal migration.

maniackcrudelis:
So, now it's tested with a small directory with spaces in sub directories.
It works as expected.

Maybe we should build a helper around that ?

maniackcrudelis:
Helper made and tested.
https://github.com/YunoHost-Apps/Experimental_helpers/commit/c8c21ef65471ecfb76ea6ac9ffbe8087bbb8c7a7

JimboJoe:
So shall this PR include and use these new helpers (I think so)...?

maniackcrudelis:
That's actually as you want, maybe it could be a good way to test and improve them.

Mickael-Martin:
Some thinks to progress to the right side :
First, my condition is wrong in "find "${datadir}/" \! -user $app -o \! -group $app -exec chown $app: {} \; ", OR condition in find work with parenthesis like that " find "${datadir}/" ! \( -user $app -o -group $app \) -exec chown $app: {} \;", so, one mistake again for me.
Second, I list all cases I try (for 15k files) : 

```
                                    If all files have wrong permissions          If all files have right permissions
Remote (find exec)                        (re-testing actually)                                      5m
Remote (chown -R)                           >30m                                                       >30m
Local (find exec)                                15s                                                            <1s
Local (chown -R)                                <1s                                                           <1s
```      
So please freeze this PR until a patch my code and my last test is finished.

JimboJoe:
Also please have a look at the aforementioned helper @maniackcrudelis just wrote that may prove useful to use/adjust so that can benefit other apps :+1:

JimboJoe:
Proposing to merge in 3 days if nobody's opposed.

maniackcrudelis:
Should we or should we not use the helpers ?

JimboJoe:
Well I think we'd better use the helpers. But seeing that @Mickael-Martin couldn't do it for the past 2 months, and provided the PR works as-is, maybe it's relevant to merge it in testing...?
@Mickael-Martin can you find some time any time soon to use the helper and apply your methodical testing?

Mickael-Martin:
Sorry for my absence. I check the helper now.

maniackcrudelis:
It fails because you add a / at the end of the directory, try with `manyFiles` instead `manyFiles/`, it works.

Mickael-Martin:
Hi, I have the same error : 
```

time ynh_chown --user=root --group=root --file="manyFiles" --recursive
+ group=root
+ recursive=1
+ '[' 1 -eq 1 ']'
+ local ending_slash=
+ '[' -d manyFiles ']'
+ ending_slash=/
+ ynh_exec_warn_less 'find "manyFiles/" \! -user root -o \! -group root | xargs -d '\''\n'\'' chown --preserve-root root:root'
+ eval find '"manyFiles/"' '\!' -user root -o '\!' -group root '|' xargs -d ''\''\n'\''' chown --preserve-root root:root
++ xargs -d '\n' chown --preserve-root root:root
++ find manyFiles/ '!' -user root -o '!' -group root
chown: missing operand after ‘root:root’
Try 'chown --help' for more information.

real    0m0.494s
user    0m0.097s
sys     0m0.319s
```
```
 xargs --show-limits
Your environment variables take up 272 bytes
POSIX upper limit on argument length (this system): 2094832
POSIX smallest allowable upper limit on argument length (all systems): 4096
Maximum length of command we could actually use: 2094560
Size of command buffer we are actually using: 131072
Maximum parallelism (--max-procs must be no greater): 2147483647
```
I ask again : "Why we don't use "-exec chown $app: {} ;" of my precedent example."

maniackcrudelis:
> I ask again : "Why we don't use "-exec chown $app: {} ;" of my precedent example."

Because unless there's a good reason, I do not change a code when it works.

Just tried again, with 14884 files, I asked for 15K with your script.
It worked perfectly, in less than 1s.

Also, there was probably a good reason to not use your code as it was. Even though I didn't keep a note about it...
However, the comment just above is probably the reason why.
> \# ' -d '\n' ' force \n to be the delimiter of each entry instead of space. So xargs will handle correctly directories and files with spaces.

alexAubin:
smol bump because this PR is pending for the testing PR :s

kay0u:
Sill waiting the use of helpers?

maniackcrudelis:
Considering that we're constantly trying to keep packages easy to maintain and read.
I'd rather prefer `ynh_chown --user=$app --file="$datadir" --recursive` than `find "$datadir/" \! -user $app -o \! -group $app | xargs -d '\n' chown $app:`...

maniackcrudelis:
Let's move on with this PR.
I've implemented the helpers for this PR, and fixed the `missing operand` error which was happening because it may occurs that there's no files on which chown or chmod have to be applied.

It works on my side, now we need a testing to have several use cases.

JimboJoe:
Looks good to me! :+1: 
@Mickael-Martin could you please try your wonderful extensive benchmarks? :slightly_smiling_face:

alexAubin:
ˢᵐᵒˡ ᵇᵘᵐᵖ :s

JimboJoe:
@Mickael-Martin any free time/will to test that PR...?

Mickael-Martin:
Successfully tested with no timeout, but a very long upgrade on my "production" (>3h) with https://github.com/Mickael-Martin/nextcloud_ynh/blob/fe95fa82f85056ea0987c80a21187ce9c7279287/scripts/upgrade#L281 (ynh_chown helper used).

JimboJoe:
So... as the performance gain is not there any more, there must be a slight difference between the helper implementation by @maniackcrudelis and the original proposal of @Mickael-Martin.
@maniackcrudelis, any opinion...?

maniackcrudelis:
Just a comparison of the commands, from the original from Mickael-Martin and the helper.
(I changed the variable names to fit the 2 commands.)

Mickael-Martin version:
`find "$datadir/" \! -user $app -o \! -group $app | xargs -d '\n' chown $app:`
Helper version:
`find "$datadir/" \! -user $app -o \! -group $app | xargs --no-run-if-empty --delimiter='\n' chown --preserve-root $app:"`

Only differences are `--no-run-if-empty` to prevent a failure of chown when there's no files to apply on. And `--preserve-root`

Mickael-Martin version:
`find "$datadir/" -type f \! -perm 0640 | xargs -d '\n' chmod 0640`
`find "$datadir/" -type d \! -perm 0750 | xargs -d '\n' chmod 0750`
Helper version:
`find "$datadir" -type f \! -perm 0640 | xargs --no-run-if-empty --delimiter='\n' chmod --preserve-root 0640"`
`find "$datadir/" -type d \! -perm 0750 | xargs --no-run-if-empty --delimiter='\n' chmod --preserve-root 0750"`
Here again, the same arguments are added. Nothing more.
We can try to remove them to try if it's faster. But I really don't see how those arguments could slower the process.

ericgaspar:
This PR was closed due to an error on my part while trying to organize branches. Please reopen it.

PR #176, opened 2019-03-10
Do not use too much process for php

maniackcrudelis:
## Problem
- *As stated by metalux on the chat room, nextcloud will crash a Raspberry Pi if it needs a lot of ram.
That's because php will spawn too much process to handle the request.*

## Solution
- *`memory_limit` wasn't of any help, since it overload the CPU instead of the ram...
So, let's limit to 3 processes instead of 10.*

## PR Status
- [x] Code finished.
- [x] Fix or enhancement tested. Crashed a raspi to do it !!! Sorry metalux :D

## Validation
---
*Minor decision*
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- **CI succeeded** : 
[Build Status](https://ci-apps-hq.yunohost.org/jenkins/job/nextcloud_ynh%20PR176/)  


JimboJoe:
I understand you got a configuration which was acceptable/usable on a RPI  (1?). Yet, what about performance settings for other targets?
According to the [official Nextcloud configuration for PHP-FPM](https://docs.nextcloud.com/server/15/admin_manual/installation/server_tuning.html#tune-php-fpm), these parameters should be higher for better performance.
So, should we discriminate based on the target hardware?

maniackcrudelis:
> pm.max_children = 120

!!! That's huge !!!
> For example on a machine with 4GB of RAM and 1GB of MySQL cache

Here I reduce to 3, which is default value for php.
But also, I did that because it was really too much for a raspi (don't know which version though). The raspi was down for a several minutes.
Using that default value would prevent a crash on small devices. Should we though use different values depending of the device ?

JimboJoe:
I've gone this way to reduce the number of processes in the Discourse app to make it bearable for an ARM device...
Here we could maybe base on NextCloudPI for ARM-based devices (see [here](https://github.com/nextcloud/nextcloudpi/blob/c5636aa240b684a2c0441fe0db160c49c2c327ea/bin/ncp/CONFIG/nc-limits.sh)...? :thinking:

maniackcrudelis:
That's also 3 for `pm.max_children` though.

But yep, using a function to define how to config that part of php would be great. For all apps by the way.
But that would mean take time to understand how it really works?
Do you think you could do something like that ?

JimboJoe:
I don't have the knowledge and I don't see myself available for such an investigation anytime soon (bye bye holidays...).
Yet, as a "best guess", maybe we could try to change directly the php-fpm helper to patch the fpm config file and set the max_children to 3 (or nproc...?) and set pm to static in case of ARM processor...?
Difficult to have a benchmark for this, though... :thinking:

maniackcrudelis:
Hard to have a benchmark indeed. That was I gave up that idea a long time ago.
But I think it would be really interesting to do such a thing.

Well, I'm not going to have so much time for the next 2 weeks, I'll have a training and I'll work during the next weekend.
But, after I will be done with the github API to rework the way ci-apps-dev gets the branches. I could have a look to that thing. I really think it could be interesting to rework that entire helper to set values according to the device, and get rid of the template we add to each app. Which is quite always the same...

maniackcrudelis:
A beginning of work here https://github.com/YunoHost-Apps/Experimental_helpers/blob/master/ynh_scalable_php/ynh_scalable_php

Eventually quite easy to work on it.
I based my work on the total ram available, the footprint of the app and the expected usage.

maniackcrudelis:
https://github.com/YunoHost-Apps/Experimental_helpers/tree/master/ynh_scalable_php
Here a full working version, with the adapted version of ynh_add_fpm_config to use either the template, like we do usually or ynh_scalable_php without any template.
Tested with leed, it works as expected.

maniackcrudelis:
git...

Closed and replaced by #247

PR #179, opened 2019-03-20
Quickfix for path traversal issue

alexAubin:
## Problem

Soooo, turns out that nextcloud is vulnerable to path traveral issues ... Going to https://domain.tld/nextcloud../html/index.nginx-debian.html might show you the nginx page... Sometime it only shows a blank white page, not sure why, but even then I can access some other dummy files (so, possibly secrets stored in other apps folders...)

Since nextcloud is one of the most used app it's a bit critical to fix this issue I guess..

## Solution

Apparently, the path traversal issues requires specific ingredient to work : 
- a `location /foo` *without* the `/` at the end
- an `alias` statement to `/var/www/foo/` *with* a `/` at the end ...

Here I simply removed the `/` at the end as it seems to be the easiest solution to fix this... But I didn't test it :/ At least the discussion is opened :stuck_out_tongue_winking_eye: :sweat_smile: 

## PR Status
- [X] Code finished.

## Validation
---
*Minor decision*
- **CI succeeded** : 
[Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20fix-path-traversal-issue%20(Official)/)  

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -6,7 +6,7 @@ location = /.well-known/caldav {
 }
 
 location ^~ __PATH__ {
-  alias __FINALPATH__/;
+  alias __FINALPATH__;
 
   if ($scheme = http) {
     rewrite ^ https://$server_name$request_uri? permanent;
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -5,8 +5,9 @@ location = /.well-known/caldav {
   return 301 https://$server_name__PATH__/remote.php/dav;
 }
 
-location ^~ __PATH__ {
-  alias __FINALPATH__;
+#sub_path_only rewrite ^__PATH__$ __PATH__/ permanent;
+location ^~ __PATH__/ {
+  alias __FINALPATH__/;
 
   if ($scheme = http) {
     rewrite ^ https://$server_name$request_uri? permanent;
@@ -42,7 +43,7 @@ location ^~ __PATH__ {
   #rewrite ^/.well-known/host-meta __PATH__/public.php?service=host-meta last;
   #rewrite ^/.well-known/host-meta.json __PATH__/public.php?service=host-meta-json last;
 
-  location __PATH__ {
+  location __PATH__/ {
     rewrite ^ __PATH__/index.php$request_uri;
   }
 


maniackcrudelis:
I've just tested https://domain.tld/nextcloud../html/index.nginx-debian.html a hour ago and it didn't work.
Just now it works !!!

That the same url !
And package check is testing `$check_domain$check_path../html/index.nginx-debian.html` which is also the same url !
There's some magic somewhere...

conf/nginx.conf
```diff
@@ -6,7 +6,7 @@ location = /.well-known/caldav {
 }
 
 location ^~ __PATH__ {
-  alias __FINALPATH__/;
+  alias __FINALPATH__;
```
maniackcrudelis: I'd rather prefer the common (and working for sure) fix, https://github.com/YunoHost/example_ynh/pull/45/files than removing this slash.

alexAubin: Yes why not, but that will also affect the `location __PATH__ {` which happens later ... so might be less trivial ... Just removing this `/` seems be more trivial (though I haven't really checked if that does work ...)

maniackcrudelis: That's the point, I'm not sure it works without consequences. Otherwise we probably wouldn't have add a rewriting rule.

alexAubin:
>There's some magic somewhere...

 A̴̬̔l̴͓̉l̶̪͝ ̷̡̆
Ć̶̼o̸̺͛m̷̱͆p̸͕͊u̵̖͂t̶̛̟e̴͈̅r̵̠̕s̴̗͛
     ̷̈͜A̷̲̾r̵̼͑e̷̘͛ ̶̨̀
  Ḇ̸́ȑ̵͖o̵͓̓ḵ̵̕ĕ̸͙n̴̜͝

conf/nginx.conf
alexAubin: Alrighty I tested it on a test server and in fact it's not clear that it really does solve the issue ... still shows a blank page when doing path traversal.

With the `/` at the end of the location, it does show a clear 404 error which sounds better

I commited the change :wink:

alexAubin: (Also changed the other `location __PATH__` in the middle to `location __PATH__/` and it seems to be okay ... Didn't do a full clean check tho)

maniackcrudelis: Meanwhile, I try to know why package_check doesn't trig an error on nextcloud

maniackcrudelis: On LXC the default file has "Welcome to nginx" as title instead of "Welcome to nginx on Debian"...

alexAubin: (╯°□°）╯︵ ┻━┻

maniackcrudelis: Also, the title changed since the time we have implement this test. Of course... It would have been too easy otherwise...

maniackcrudelis:
Looks ok

JimboJoe:
Will be merged in 3 days.

alexAubin:
Shall we merge it in testing after https://github.com/YunoHost-Apps/nextcloud_ynh/pull/181 ?

PR #237, opened 2019-10-25
Urgent security issue in NGINX/php-fpm

ZeJMaN:
more information here  : https://nextcloud.com/blog/urgent-security-issue-in-nginx-php-fpm/

## Problem
In the last 24 hours, a new security risk has emerged around NGINX, documented in CVE-2019-11043. This exploit allows for remote code execution on some NGINX and php-fpm configurations

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -47,7 +47,7 @@ location ^~ __PATH__/ {
   #rewrite ^/.well-known/host-meta.json __PATH__/public.php?service=host-meta-json last;
 
   location __PATH__/ {
-    rewrite ^ __PATH__/index.php$request_uri;
+    rewrite ^ __PATH__/index.php;
   }
 
   location = __PATH__/robots.txt {
@@ -66,6 +66,7 @@ location ^~ __PATH__/ {
   location ~ ^__PATH__/(?:index|remote|public|cron|core/ajax/update|status|ocs/v[12]|updater/.+|ocs-provider/.+|ocm-provider/.+)\.php(/.*|)$ {
     include fastcgi_params;
     fastcgi_split_path_info ^(.+\.php)(/.+)$;
+    try_files $fastcgi_script_name =404;
     fastcgi_param SCRIPT_FILENAME $request_filename;
     fastcgi_param PATH_INFO $fastcgi_path_info;
     fastcgi_param HTTPS on;


conf/nginx.conf
```diff
@@ -47,7 +47,7 @@ location ^~ __PATH__/ {
   #rewrite ^/.well-known/host-meta.json __PATH__/public.php?service=host-meta-json last;
 
   location __PATH__/ {
-    rewrite ^ __PATH__/index.php$request_uri;
+    rewrite ^ __PATH__/index.php;
```
alexAubin: Thanks for the patch :+1: 

Do you know if that specific change will work without upgrading the php version ?

ZeJMaN: i just apply this change on my production server which runs in stable version of yunohost
For information here is the version of php that I use with this modification:

PHP 7.0.33-0+deb9u5 (cli) (built: Sep 18 2019 09:55:34) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.33-0+deb9u5, Copyright (c) 1999-2017, by Zend Technologies

JimboJoe:
Thanks for that!
What about the second part about adding `try_files $fastcgi_script_name =404;`?

JimboJoe:
Shouldn't we accelerate the process and merge directly to master without waiting for 3 days, after the second approval...?

kay0u:
> Shouldn't we accelerate the process and merge directly to master without waiting for 3 days, after the second approval...?

I don't know, it would be great to speed up the process, but I'm not for merging directly into master. It could be merged into testing, and merge soon the Testing branch, what do you think about that?

JimboJoe:
Well, why not, but there's already a pending testing, so that people will be forced to upgrade to 15.0.12 on a very short term...

kay0u:
Not false (even if I personally think it's better to have the latest version, some people might not think like me), then you have to upgrade the package version, and merge master into testing right after.

We can say let's merge in 1 or 2 days in master (to let other people of app group give their point of view?)

kay0u:
Or now if you think it's too risky to do this in a day or two?

JimboJoe:
Well, 15.0.12 testing has been out for a while now, so let's assume it's OK to release any time. So maybe:
* merge this patch to testing by the end of the day, to let any other one from the app group step in, and advertise the testing (15.0.12 + security patch)
* release the stable version in a couple of days.

What about that?

kay0u:
I'm ok!

lapineige:
I'd personally prefer a merge in master directly (and ASAP), regarding the magnitude of the security issue. Considering the time to advertise the update (and for people to get the information), and the *relative* difficulty for non-technical people to update to a testing branch, even a few days can counts.

Yet the current solution seems reasonable 🙂 

Btw, this branch works perfectly fine on 2 servers of mine.

mthld:
Hi,

> I'd personally prefer a merge in master directly (and ASAP), regarding the magnitude of the security issue. Considering the time to advertise the update (and for people to get the information), and the _relative_ difficulty for non-technical people to update to a testing branch, even a few days can counts.

I second this, the changes made in this PR aren't subsequents: this is only small Nginx parameters. They can't harm an instance in any way.

zamentur:
LGTM, i think we should release this fix earlier.

zamentur:
Here a small testimony about an attach that succeed: https://forum.chatons.org/t/nextcloud-nginx-fpm-update-required/577/5

JimboJoe:
OK, so shall we merge this in testing and release right away?
We need a clear and quick opinion from everyone!
I'd say I'm in favour of releasing 15.0.12 right away.

lapineige:
I don't know what's best, but any solution that will release the fix right away is what we need.

zamentur:
I agree with merging this **evening** into master with or without testing change, as you want.

lapineige:
And advertise it at the same time.

lapineige:
> What about the second part about adding try_files $fastcgi_script_name =404;?

🆙

PR #260, opened 2020-03-14
Testing

maniackcrudelis:
## Problem
- *Version 15 is not supported anymore.*
- *The onlyoffice builtin feature in 18 is very useful.*
- *Install fails on buster because there's no php-mcrypt anymore*

## Solution
- *Add some php dependencies as suggested in https://docs.nextcloud.com/server/18/admin_manual/installation/source_installation.html to make some features working*
- Fix https://github.com/YunoHost-Apps/nextcloud_ynh/issues/257 https://github.com/YunoHost-Apps/nextcloud_ynh/issues/256 https://github.com/YunoHost-Apps/nextcloud_ynh/issues/205
- *Use php7.3*
- *Fix php issues with php7.4*
- *Fix the memory limit*
- *Fix detection of another caldav/carddav server*
- *Upgrade to last 18.0.2*

After a quick lookup on the internet, my understanding is that the stuff provided by php-mcrypt is old / deprecated since a while, and in fact Nextcloud doesnt need it anymore (it was added back to the jessie or maybe even squeeze era).

So let's remove it. (N.B. : I haven't tested anything though - but it should be okay to commit this for stretch as well)

## PR Status
- [x] Code finished.
- [x] Tested with Package_check.
- [x] Fix or enhancement tested.
- [x] Upgrade from last version tested.
- [x] Can be reviewed and tested.

## Validation
---
*Minor decision*
- **Upgrade previous version** : 
- [x] **Code review** : Maniack C
- [x] **Approval (LGTM)** : Maniack C
- [x] **Approval (LGTM)** : Kay0u
- **CI succeeded** : 
[Build Status](https://ci-apps-hq.yunohost.org/jenkins/job/nextcloud_ynh%20PR260/)  
When the PR is marked as ready to merge, you have to wait for 3 days before really merging it.

PR #264, opened 2020-03-21
Don't use parentheses

kay0u:
## Problem
- *Parentheses are useless here (I guess?)*
- *Parentheses create a subshell, and ynh_die exit only the subshell, not the main script*

## Solution
- *Remove parentheses*

## PR Status
- [ ] Code finished.

## Validation
---
*Minor decision*
- **CI succeeded** : 
[Build Status](https://ci-apps-hq.yunohost.org/jenkins/job/nextcloud_ynh%20PR264/)  

### DIFF
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -292,7 +292,7 @@ then
         # Upgrade Nextcloud (SUCCESS = 0, UP_TO_DATE = 3)
         exec_occ maintenance:mode --off
         exec_occ upgrade \
-        || ([ $? -eq 3 ] || ynh_die --message="Unable to upgrade Nextcloud")
+        || [ $? -eq 3 ] || ynh_die --message="Unable to upgrade Nextcloud"
 
         # Get the new current version number
         current_version=$(grep OC_VersionString "$final_path/version.php" | cut -d\' -f2)


JimboJoe:
I'm sorry, can you please precise the rationale in the PR description... ? :innocent:

kay0u:
Done :-)

maniackcrudelis:
For sure the parenthesis break ynh_die, which will exit the sub shell only. But are you sure about the priority of the logical OR ?
Did you actually test the behavior of this ?

kay0u:
I have to admit that I didn't test this code, but in bash leftmost operations are executed first

https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Lists

Anyway I'll run some tests

kay0u:
Should we merge it in 3 days or when #260 is merged?

maniackcrudelis:
That doesn't look like a hotfix, so I guess it can wait until the next release.

alexAubin:
By next release you mean the current testing ? To me it *does* look like a kinda urgent thing to merge considering that if the exec_occ upgrade fails, the script will continue and it's likely to create a huge mess (though I didn't take a look at what happens next). So at least should be merged for the 18.x considering many people are going to upgrade ?

maniackcrudelis:
Considering that error in the code is 4 years old, I don't think we should really consider it like a big issue !
This error appear in this https://github.com/YunoHost-Apps/nextcloud_ynh/commit/5cbd01ab57c8550ca6d60538ff87d55493d40a58#diff-44cb16c778719320333118c04d509a7cR98...

kay0u:
On the other hand, since this is a fairly major upgrade, it would be best to fix any known issue to avoid a shit storm...

lapineige:
I don't know the details and implications, but if it creates no major additional risk during the upgrade, I think too that it's good to merge that with the rest.
Also, people might not upgrade their Nextcloud so often (considering it's not a "simple" app, and as it broke a few time in the past, for my part I try to be cautious with its updates), and this would delay the fix maybe until the next major version maybe.

maniackcrudelis:
To be honest, I don't use Nextcloud so much and my version 15 works pretty well for what I need to do.

But, I try to not postpone any more this release everyone seem to wait for.
So that I try to not include any new commit that are not really important.

Considering this code works since 4 years. I'm more worried about what will happen when merged than the opposite.

Still, there's 2 solutions, either we merge right now into testing and we'll see what happen in the next 3 days. In hope a regression wouldn't go unnoticed.
Or we merge in 3 days and postpone again the release of this testing.

Again, I won't probably jump on the occasion to upgrade my own nextcloud to 18. So, both solutions fit me. If you really don't want to wait for the next release of course.

lapineige:
If it's a matter of 3 more days with a testing branch with this fix and a master branch with v18, maybe it's not a big deal (by the time people learn about the available upgrade + actually upgrade their instance). Yet that might mean a lot of upgrades, as v18 is much expected.
FYI I'm gonna test it right now.

PR #266, opened 2020-03-24
fix caldav/carddav detection

kay0u:
## Problem
- *La configuration du serveur web ne permet pas d'atteindre "/.well-known/caldav". Vous trouverez plus d'informations dans la documentation.*
- *La configuration du serveur web ne permet pas d'atteindre "/.well-known/carddav". Vous trouverez plus d'informations dans la documentation.*

## Solution
- *Reload nginx before check url*

## PR Status
- [x] Code finished.

## Validation
---
*Minor decision*
- [x] **Code review** : Maniack C
- **CI succeeded** : 
[Build Status](https://ci-apps-hq.yunohost.org/jenkins/job/nextcloud_ynh%20PR266/)  

### DIFF
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -146,7 +146,7 @@ ynh_install_php --phpversion="$YNH_PHP_VERSION" --package="$extra_pkg_dependenci
 #=================================================
 # NGINX CONFIGURATION
 #=================================================
-ynh_script_progression --message="Upgrading nginx web server configuration..." --weight=3
+ynh_script_progression --message="Upgrading nginx web server configuration..." --weight=2
 
 ynh_backup_if_checksum_is_different --file="/etc/nginx/conf.d/$domain.d/$app.conf"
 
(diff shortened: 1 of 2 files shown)


maniackcrudelis:
Into the install script this helper is used before the reload of nginx.
I see that into the upgrade script, it's after the config is removed, with a reload of nginx.

I don't think that sleep should be into the helper, if really needed, into the upgrade script.

kay0u:
+1
I tested without the sleep, it failed one time, it failed once, but now it's working, so I prefer to leave it to be sure.

scripts/upgrade
```diff
@@ -146,14 +146,19 @@ ynh_install_php --phpversion="$YNH_PHP_VERSION" --package="$extra_pkg_dependenci
...
+ynh_script_progression --message="Upgrading nginx web server configuration..." --weight=3
 
 ynh_backup_if_checksum_is_different --file="/etc/nginx/conf.d/$domain.d/$app.conf"
 
 # Delete current nginx configuration to be able to check if .well-known is already served.
-ynh_remove_nginx_config $YNH_PHP_VERSION
+ynh_remove_nginx_config
 ynh_app_setting_delete --app=$app --key="checksum__etc_nginx_conf.d_$domain.d_$app.conf"
 
+ynh_systemd_action --service_name=nginx --action=reload
```
maniackcrudelis: There's already a reload into the helper.
If you really want to add a reload here, use the log detection in replacement of the sleep.

kay0u: Ho sorry I forgot about that. Maybe just a sleep then to be sure? I can't reproduce 100% the issue --'

maniackcrudelis: I guess 1s isn't that much, but using a sleep is hardly a reliable fix

kay0u: You're right, I did this quick and dirty fix because we have the same here:
https://github.com/YunoHost/yunohost/blob/00a9641e1458b92c4cb099da6dfb9e38bb35e82d/data/helpers.d/utils#L239

maniackcrudelis: Which wasn't a reliable way to fix it either...
Anyway, if you try your last fix and if it indeed works, I love this one :)

kay0u: Yep I try to install/upgrade with and without this fix, it seems to work™

JimboJoe:
Can be merged in 3 days.

PR #282, opened 2020-04-01
[fix] Backward compatibility for NC 14 or older

zamentur:
## Problem
Can't upgrade from nextcloud 13 (and probably 14)

## Solution
Use php7.0 for upgrade these version and next use the new php version

## PR Status
- [x] Code finished.

## Validation
---
*Minor decision*
- [x] **Code review** : Maniack C
- **CI succeeded** : 
[Build Status](https://ci-apps-hq.yunohost.org/jenkins/job/nextcloud_ynh%20PR282/)  

### DIFF
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -200,11 +200,11 @@ fi
 # Define a function to execute commands with `occ`
 exec_occ() {
     # Backward compatibility to upgrade from NC14 or older version
-    if [ $current_major_version -gt 14 ]
+    if [ $current_major_version -lt 15 ]
     then
-	NEXTCLOUD_PHP_VERSION=$YNH_PHP_VERSION
-    else
         NEXTCLOUD_PHP_VERSION="7.0"
+    else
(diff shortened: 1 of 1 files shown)


scripts/upgrade
```diff
@@ -199,8 +199,15 @@ fi
 
 # Define a function to execute commands with `occ`
 exec_occ() {
+    # Backward compatibility to upgrade from NC14 or older version
+    if [ $current_major_version -lt 15 ]
+    then
+        NEXTCLOUD_PHP_VERSION="7.0"
```
kay0u, line 205: It's going to fail on buster, do we have to deal with it? (and maybe install php 7.0 for the migration in this case?)

kay0u:
Can be merged in 3 days

scripts/upgrade
maniackcrudelis: Considering there's people on jessie...
I think that's going to be a problem sooner or later...

Maybe in such situation we could consider a manual way around. Anyway, if that happen, we will know.

PR #305, opened 2020-05-27
Update nginx

kay0u:
## Problem
- *Manual modification was required to use onlyoffice*

## Solution
- *Apply [example nginx configuration](https://docs.nextcloud.com/server/18/admin_manual/installation/nginx.html)*
- *rewrite `documentserver_community` with `index.php` to be able to use onlyoffice right after installing it*

## PR Status
- [ ] Code finished.

## Validation
---
*Minor decision*
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- **CI succeeded** : 
[Build Status](https://ci-apps-hq.yunohost.org/jenkins/job/nextcloud_ynh%20PR305/)  

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -69,15 +69,15 @@ location ^~ __PATH__/ {
     access_log off;
   }
 
-  location ~ ^__PATH__/(?:build|tests|config|lib|3rdparty|templates|data)/ {
+  location ~ ^\__PATH__\/(?:build|tests|config|lib|3rdparty|templates|data)\/ {
     deny all;
   }
-  location ~ ^__PATH__/(?:\.|autotest|occ|issue|indie|db_|console) {
+  location ~ ^\__PATH__\/(?:\.|autotest|occ|issue|indie|db_|console) {
     deny all;
   }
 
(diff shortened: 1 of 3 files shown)


JimboJoe:
Let's merge in 3 days.

conf/nginx.conf
```diff
@@ -69,15 +69,15 @@ location ^~ __PATH__/ {
     access_log off;
   }
 
-  location ~ ^__PATH__/(?:build|tests|config|lib|3rdparty|templates|data)/ {
+  location ~ ^\__PATH__\/(?:build|tests|config|lib|3rdparty|templates|data)\/ {
```
alexAubin, line 72: I'm randomly jumping in but won't this trigger weird issues considering that the nginx helper replaces `__PATH__` and `__PATH__/` by different thing in the case of installation on a domain root ? Or is it in fact on purpose ?

JimboJoe: but kay0u it was tested... was it? :sweat_smile:

kay0u: https://github.com/YunoHost/yunohost/blob/051d14a945ef00317cf79061389fb15b0fee951b/data/helpers.d/nginx#L35-L36

What kind of different thing? I tested it, it works, but I have to admit that I just copy pasta the current example, and I'm not sure if we should keep these `\`

Issue #331, opened 2020-09-06
Element (Riotchat) doesn't work


Thatoo:
Everything is explained in this thread : https://github.com/gary-kim/riotchat/issues/26

Installations works without problem, but when we go to the app, nothing appears.

Someone who tried to dig the issue said : 

> So, the issue seems to be that when loading Riot, the requests for the static resources are being redirected to the login page by Yunohost. I don't know why that is but it seems like it's happening to quite a few of the requests. I think this would be an upstream issue in Yunohost. Gonna try debugging a little further.

kay0u:
Hi, this PR should explain the problem, and propose a solution: https://github.com/YunoHost-Apps/nextcloud_ynh/pull/332

Thatoo:
A solution has been given to solve a problem that seemed to me to be similar (error with keeweb in nextcloud) The solution is here https://github.com/YunoHost-Apps/nextcloud_ynh/issues/317#issuecomment-702596438

However, even with this solution, Element is not working. So the problem is different, it might be indeed the one pointed by @kay0u . Did someone try to disable pretty url and tested Element in Nextcloud? Is it working?

kay0u:
Should be fixed now

PR #332, opened 2020-09-08
Disable pretty urls

kay0u:
## Problem

The default nginx configuration uses `root` rather than `alias`. This works very well, but from what I understood from my tests there are some limitations:
According to the [doc](https://docs.nextcloud.com/server/latest/admin_manual/installation/nginx.html):
- For nextcloud installed at the root of a domain, we use `root /var/www/nextcloud;`, everything works with pretty urls.
- For nextcloud installed on a subpath, then we use `root /var/www;` instead. Which is a problem for us because then nextcloud would have to be in the `/var/www/url_path_chosen_by_the_user` directory.

So we have to replace `root` by `alias` (which works very well), but then with pretty urls enabled, many apps don't work anymore, like for example: https://github.com/YunoHost-Apps/nextcloud_ynh/issues/331, `documentserver_community` where we had to add a specific `location`... Anyway, I searched for a long time, and I couldn't find any other solution than to disable it.

We will then have urls of the form:

`https://domain.tld/nextcloud/index.php/apps/....`
instead of
`https://domain.tld/nextcloud/apps/....`

How bad is it?

## Solution
- *Disable pretty urls*

## PR Status
- [ ] Code finished.

## Validation
---
*Minor decision*
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- **CI succeeded** : 
[Build Status](https://ci-apps-hq.yunohost.org/jenkins/job/nextcloud_ynh%20PR-NUM-/)  
*Please replace '-NUM-' in this link by the PR number.*  

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -90,7 +90,9 @@ location ^~ __PATH__/ {
     # Avoid sending the security headers twice
     fastcgi_param modHeadersAvailable true;
     # Enable pretty urls
-    fastcgi_param front_controller_active true;
+    # Do not enable pretty urls because it break a lot of apps
+    # To make it works, we must change "alias" into "root", which creates another set of problems.
+    #fastcgi_param front_controller_active true;
     fastcgi_pass unix:/var/run/php/php__YNH_PHP_VERSION__-fpm-__NAME__.sock;
     fastcgi_intercept_errors on;
     fastcgi_request_buffering off;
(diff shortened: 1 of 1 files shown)


zamentur:
it means all shared link will be broken, so it's quite bad...

kay0u:
> it means all shared links and cardav/calldav/webdav config will be broken, so it's quite bad...

After some tests, it seems that the shared links on files are not broken nor cardav/calldav/webdav, but shared links on a calendar are broken.

I don't think we can tests every use case,  but I can maybe disable pretty urls on fresh install but keep them on the old ones?

Thatoo:
what means "shared links on a calendar"?
If we shared a link for someone to be able to have a look to one specific calendar?

> I don't think we can tests every use case, but I can maybe disable pretty urls on fresh install but keep them on the old ones?

Will it be possible for old one to disable pretty url too if we prefer to be able to use apps such as element and keeweb and don't mind to loose some "shared links on a calendar" whatever that mean?

zamentur:
How 

> After some tests, it seems that the shared links on files are not broken nor cardav/calldav/webdav, but shared links on a calendar are broken.

I don't understand how it could be possible...

Example: 
i have currently a nextcloud 19 with this share link https://example.com/fichier/s/JJ4dSR9ngGkrXiA

- I give this link by email to a friend
- I do the upgrade
- My friend try to access https://example.com/fichier/s/JJ4dSR9ngGkrXiA
- BUT it could not work because pretty url are not anymore supported and the person needs to access via https://example.com/fichier/index.php/s/JJ4dSR9ngGkrXiA

Maybe i don't understand what this PR does.

zamentur:
> what means "shared links on a calendar"?
> If we shared a link for someone to be able to have a look to one specific calendar?

Yes, in nextcloud you have 2 kind of share, you can share with an other nextcloud account, or you can share by link. It could be done on file, on calendar, etc.

sxenny:
hi guess it seem like solutions for my problem with an empty Element(Riochat) screen. How I can disable pretty URL for yunohost nextcloud? (or all yunohost)

kay0u:
Closed by https://github.com/YunoHost-Apps/nextcloud_ynh/pull/346

PR #340, opened 2020-10-07
Fix issues with Element (Riotchat) app

sebmennetrier:
## Problem
- See bug #331

## Solution
- Redirect call to riot files using `location...` to use something like `https://domain.tld/nextcloud/index.php/apps/....`
This specific configuration is similar to the conf of `documentserver_community`

## PR Status
- [x] Code finished.

## Validation
---
*Minor decision*
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- **CI succeeded** : 
[Build Status](https://ci-apps-hq.yunohost.org/jenkins/job/nextcloud_ynh%20PR340/)

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -121,6 +121,10 @@ location ^~ __PATH__/ {
     rewrite ^ __PATH__/index.php$request_uri;
   }
 
+  location ~* ^__PATH__/apps/riotchat/riot/ {
+    rewrite ^ __PATH__/index.php$request_uri;
+  }
+
   # Adding the cache control header for js, css and map files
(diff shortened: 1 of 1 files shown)


Thatoo:
It does work!
Awesome.
No need of removing pretty url then.

Thatoo (approved):

kay0u:
Some other apps may have the same problem, but let's merge this into testing

kay0u (approved):

sebmennetrier:
Thanks !!

I totally agree. To fix this type of issue today, we can remove pretty links that seems a little complicated or we can add this type of exception for every applications that needed. The number of application seems limited at the moment so maybe that this type of patch can be a solution waiting a better 😉

PR #346, opened 2020-10-26
Testing

kay0u:
## Problem
- *New version upstream*
- *See bug #331*

## Solution
- *Upgrade to version 20.0.4*
- *Small typos*
- *Rework the nginx conf (close #331 close #332 close #317)*
- *Allow Symlinks (Fix #345)*

## PR Status
- [ ] Code finished.
- [ ] Tested with Package_check.
- [ ] Fix or enhancement tested.
- [ ] Upgrade from last version tested.
- [ ] Can be reviewed and tested.

## Validation
---
*Minor decision*
- **Upgrade previous version** : 
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- **CI succeeded** : 
[Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20PR346%20(Kay0u)/)  [pipeline status](https://git.kayou.io/kayou/nextcloud_ynh/-/commits/testing) 
When the PR is marked as ready to merge, you have to wait for 3 days before really merging it.

PR #360, opened 2020-12-29
disable mail app to fix the upgrade from 19 to 20

kay0u:
## Problem

https://github.com/YunoHost-Apps/nextcloud_ynh/pull/346#issuecomment-751996938
> Upgrade unfortunately still fails if the mail app is installed and enabled...

## Solution
- *Disable mail app during the upgrade*

## PR Status
- [ ] Code finished.

## Validation
---
*Minor decision*
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- **CI succeeded** : 
[Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20PR360%20(Kay0u)/)  

### DIFF
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -245,7 +245,8 @@ then
     exec_occ -V
 
     # Upgrade may fail if this app is enabled
-    exec_occ app:list | grep -q -w mail
+    # Take all apps enabled, and check if mail is one of them
+    exec_occ app:list | awk '/Enabled/{f=1;next} /Disabled/{f=0} f' | grep -q -w mail
     mail_app_is_active=$?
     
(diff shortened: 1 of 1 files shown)


kay0u:
Yolomerging

Bugsbane:
Might I suggest that if not having 4 byte support makes upgrades fail when Mail / News are installed, that instead of disabling Mail/News apps, what about if the database is upgraded to allow 4 byte support instead? That way all functionality remains intact, unless I'm missing something... If there are edge cases where that may not work, disabling apps could always be the fallback behavior...

PR #377, opened 2021-02-19
Upgrade to version 21.0.0

ericgaspar:
## Problem
- *New version upstream*

## Solution
- *Upgrade to version 21.0.0*

## PR Status
- [ ] Code finished.
- [ ] Tested with Package_check.
- [ ] Fix or enhancement tested.
- [ ] Upgrade from last version tested.
- [ ] Can be reviewed and tested.

## Validation
---
*Minor decision*
- **Upgrade previous version** : 
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 

## Package_check results
---
* An automatic package_check will be launch at https://ci-apps-dev.yunohost.org/, when you add a specific comment to your Pull Request: "!testme", "!gogogadgetoci" or "By the power of systemd, I invoke The Great App CI to test this Pull Request!"*

PR #382, opened 2021-02-23
Fixes Nextcloud 21

Thovi98:
## Problem
https://github.com/YunoHost-Apps/nextcloud_ynh/pull/377#issuecomment-781999502
`/.well-known/webfinger` and `/.well-known/nodeinfo` issue in Nextcloud 21

`Votre installation n’a pas de préfixe de région par défaut. C’est nécessaire pour valider les numéros de téléphone dans les paramètres du profil sans code pays. Pour autoriser les numéros sans code pays, veuillez ajouter "default_phone_region" avec le code ISO 3166-1 respectif de la région dans votre fichier de configuration.`

## Solution
- update nginx as proposed in the docs (https://docs.nextcloud.com/server/21/admin_manual/installation/nginx.html)

- for default_phone_region issue, we have to configure it in `config/config.php`, but I don’t know which one set as default, because there are different country users who install this package... (https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#user-experience)
`
## PR Status
- [x] Code finished.

---

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -14,7 +14,7 @@ location ^~ /.well-known {
   location = /.well-known/caldav      { return 301 __PATH__/remote.php/dav/; }
  
   # Anything else is dynamically handled by Nextcloud
-  location ^~ /.well-known          { return 301 /nextcloud/index.php$uri; }
+  location ^~ /.well-known          { return 301 __PATH__/index.php$uri; }
 
   try_files $uri $uri/ =404;
 }
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -12,6 +12,9 @@ location ^~ /.well-known {
 
   location = /.well-known/carddav     { return 301 __PATH__/remote.php/dav/; }
   location = /.well-known/caldav      { return 301 __PATH__/remote.php/dav/; }
+ 
+  # Anything else is dynamically handled by Nextcloud
+  location ^~ /.well-known          { return 301 /nextcloud/index.php$uri; }
 
   try_files $uri $uri/ =404;
 }


conf/nginx.conf
```diff
@@ -12,6 +12,9 @@ location ^~ /.well-known {
 
   location = /.well-known/carddav     { return 301 __PATH__/remote.php/dav/; }
   location = /.well-known/caldav      { return 301 __PATH__/remote.php/dav/; }
+ 
+  # Anything else is dynamically handled by Nextcloud
+  location ^~ /.well-known          { return 301 /nextcloud/index.php$uri; }
```
kay0u, line 17:
```suggestion
  location ^~ /.well-known          { return 301 __PATH__/index.php$uri; }
```

But, the problem with this line is if another app which uses `.well-known/something` is installed on the same domain, then nextcloud will handle it instead of this app

Thovi98: Okay, I will look further and see if I find something (I know nothing about all this, so don’t expect a lot form me 😄)

kay0u: A workaround we found is the lines above: https://github.com/YunoHost-Apps/nextcloud_ynh/blob/ee18542f4bc7dd888e7f8bfbae92d9e076c27ccb/conf/nginx.conf#L6-L11
This requires to edit the file manually for users, which is not really YunoHost-friendly

Thovi98: And adding only 
```
 location = /.well-known/webfinger   { return 301  __PATH__/index.php$uri; }
 location = /.well-known/nodeinfo   { return 301 __PATH__/index.php$uri; }
```

is not a good alternative as well ?

kay0u: Yes probably :+1: 

After some tests, I can confirm it works. Let's merge it and I'll patch nginx conf

PR #125, opened 2018-06-17
Use PostgreSQL instead of MySQL (fixes #107)

JimboJoe:
## Problem
- *MySQL is much less robust than PostgreSQL*

## Solution
- *Migrate Nextcloud to PostgreSQL*

## PR Status
- [ ] Code finished.
- [ ] Tested with Package_check.
- [ ] Fix or enhancement tested.
- [ ] Upgrade from last version tested.
- [ ] Can be reviewed and tested.

## Validation
---
*Minor decision*
- **Upgrade previous version** : 
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 
- **CI succeeded** : 
[Build Status/badge/icon)](https://ci-apps-dev.yunohost.org/jenkins/job/nextcloud_ynh%20enh_postgresql_migration%20(Official)/)

 [pipeline status](https://git.kayou.io/kayou/nextcloud_ynh/-/jobs/444) 

When the PR is marked as ready to merge, you have to wait for 3 days before really merging it.

PR #384, opened 2021-02-25
Migrate mysql db to utf8

kay0u:
## Problem
- *Description of why you made this PR*

## Solution
- *Fix https://github.com/YunoHost-Apps/nextcloud_ynh/pull/384*

## PR Status
- [ ] Code finished.

## Validation
---
*Minor decision*
- [ ] **Code review** : 
- [ ] **Approval (LGTM)** : 
- [ ] **Approval (LGTM)** : 

## Package_check results
---


kay0u:
!testme

Samuel-BF:
While you're at it, I would suggest to change charset to utf8mb4 : utf8 [is an alias for utf8mb3](https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8.html), which is the UTF-8 subset of characters that are at most 3 bytes long. utf8mb4, on the other side, cant represent any UTF-8 value.

scripts/install
```diff
@@ -66,6 +66,9 @@ db_name=$(ynh_sanitize_dbid --db_name=$app)
 ynh_app_setting_set --app=$app --key=db_name --value=$db_name
 ynh_mysql_setup_db --db_user=$db_name --db_name=$db_name
 
+ynh_mysql_connect_as --user=$db_name --password="$db_pwd" --database=$db_name \
+    <<< "ALTER DATABASE $db_name CHARACTER SET utf8 COLLATE utf8_general_ci;"
```
kay0u, line 70:
```suggestion
    <<< "ALTER DATABASE $db_name CHARACTER SET  utf8mb4 COLLATE utf8_general_ci;"
```

kay0u:
!testme

kay0u:
Thank you for pointing this out to me, let see if it works

kay0u:
Hmm it doesn't work

kay0u:
!testme

kay0u:
!testme

kay0u:
!testme

Thatoo:
I'm not sure to understand... @alexAubin Is it possible to migrate the database to utf8mb4? Will be done during migration do bullseyes?

lapineige:
Hello,

Any news about that ? I might be running into issues because of that non-UT8 character set for another app (FreshRSS https://github.com/FreshRSS/FreshRSS/issues/4139) and I wonder if that's a change I need to do for all databases or if they are independents in some way.

Thanks in advance :)

kay0u:
I think this PR will be obsolete after the Bullseye migration if I understood correctly.

alexAubin:
Eeeeh not exactly because the new config only affects *newly create tables*. For old one, one still has to convert the content of the table, which is not a trivial operation and is apparently app-specific (you can't just take any random SQL DB and convert it). This is why the `exec_occ db:convert-mysql-charset` exists and handles the reencoding.

kay0u:
!testme

ericgaspar:
!testme

Thatoo:
Does that mean old yunohost nextcloud won't be able to migrate mysql to utf8?

ericgaspar:
!testme

Thatoo:
I'm a bit confused. Do we need to migrate mysql to utf8 if we migrate to PostgreSQL (https://github.com/YunoHost-Apps/nextcloud_ynh/pull/125) ?

Thatoo:
Is it working now with `maintenance:repair` ? Why can't we merge it into testing?

ericgaspar:
!testme

MCMic:
Hello,

If I fix this by hand, will that cause trouble later on application updates?

Otherwise, are there blockers left to merging this? It blocks news application as soon as a feed uses special characters.

ericgaspar:
> Otherwise, are there blockers left to merging this? It blocks news application as soon as a feed uses special characters.

We need user testing this branch and feedback I guess.

lapineige:
Let's merge to testing once #532 is merged ? Hence this might attract more attention.

If we are wiling to test, what can we do to test it ?

ericgaspar:
!testme

Limezy:
FYI I've got a problem now on my instance for which I have run this PR.
I'm still investigating but I don't think it's safe to merge yet

ericgaspar:
can you be more specific?

Limezy:
Sorry, after investigating the problem is *NOT* related (I'm 100% sure).
I had done the postgresql upgrade just after that upgrade, using the -b option, and I forgot it, so I thought that after the failure the backup was restored, but it wasn't. So I had a nextcloud app showing, but with the entire mysql table dropped 🤣

lapineige:
Just to make sure, this will be updated automatically, nothing to do on our side ?

PR #394, opened 2021-03-17
Make sure to redact admin password

alexAubin:
## Problem

- Password appearing in clear in logs

## Solution

- The password was not redacted by Yunohost auto-redact mecanism because it looks for patterns like `[something]password=[password]` (like, you put the password in a variable to be used later)

## PR Status

Yolocommited

## Package_check results
---

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -147,11 +147,16 @@ exec_occ() {
 # Set write access for the following commands
 chown -R $app: "$final_path" "$datadir"
 
+# Define password in an intermediate var
+# The fact that it's called _password allows it to be
+# picked up by Yunohost's auto-redact mecanism
+admin_password="$(ynh_string_random --length=6)"
+
 # Install Nextcloud using a temporary admin user
 exec_occ maintenance:install \
     --database "mysql" --database-name $db_name \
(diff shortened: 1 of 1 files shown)


scripts/install
```diff
@@ -147,11 +147,16 @@ exec_occ() {
 # Set write access for the following commands
 chown -R $app: "$final_path" "$datadir"
 
+# Define password in an intermediate var
+# The fact that it's called _password allows it to be
+# picked up by Yunohost's auto-redact mecanism
+admin_password="$(ynh_string_random --length=6)"
```
alexAubin, line 153: By the way length=6 ... that doesn't sounds super secure x_x

kay0u: Yes, but we remove this user later: https://github.com/YunoHost-Apps/nextcloud_ynh/blob/41d2c377623d58862c8ee3dc2dc70f1f8351fe4f/scripts/install#L258

kay0u (approved):

scripts/install
alexAubin: Oh I see

So less spooky than I tought, nvm

PR #455, opened 2021-11-14
Add --purge option

ericgaspar:
- Add `--purge option`to remove datadir

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -130,9 +130,12 @@ ynh_add_nginx_config
 #=================================================
 # CREATE THE DATA DIRECTORY
 #=================================================
+ynh_script_progression --message="Creating a data directory..." --weight=1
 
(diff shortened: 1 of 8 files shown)


ericgaspar:
!testme

scripts/install
```diff
@@ -130,9 +130,12 @@ ynh_add_nginx_config
 #=================================================
 # CREATE THE DATA DIRECTORY
 #=================================================
+ynh_script_progression --message="Creating a data directory..." --weight=1
 
 # Define app's data directory
 datadir="/home/yunohost.app/$app/data"
+ynh_app_setting_set --app=$app --key=datadir --value=$datadir
```
kay0u, line 137: You should manage when the datadir setting is empty in the upgrade script

PR #466, opened 2021-12-08
Update check_process

ericgaspar:
- Set check process for config panel

### DIFF
--- a/check_process
+++ b/check_process
@@ -17,7 +17,7 @@
 		setup_private=1
 		setup_public=1
 		upgrade=1
-		#21.0.2 
+		#22.2.0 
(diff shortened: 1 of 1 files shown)


ericgaspar:
!testme

ericgaspar:
@alexAubin Are the new config-panels testable with package-check?

kay0u:
Hmm I'd say no

alexAubin:
Hmpf yeah package_check needs to be reworked to interface with the new config panel API ... And maybe we need to think what / how we want to test stuff exactly. For example it's quite boring to have to declare all that stuff in check_process. Also in the general case there are more complex panels than just single independent questions (c.f. for example vpnclient, hotspot) so it's not clear how to test those.

Issue #484, opened 2022-04-28
JSON requests wrongfully cached


nicofrand:
### Describe the bug

Actions in nextcloud's app [Cospend](https://github.com/eneiluj/cospend-nc) are not updating the UI.
To sum up, creating or deleting a bill in Cospend then triggers a request to refresh the interface's data.
This request does not return the right (updated) data but a cached one, therefore invalid.

![image](https://user-images.githubusercontent.com/3419050/165721897-ce1ddd17-5357-4217-869a-343e05e9dc41.png)

URL called: https://mydomain.tld/apps/cospend/projects/courses

### Context

- Nextcloud installed through YunoHost
- Cospend installed through Nextcloud's webadmin

### Steps to reproduce

- Install Nextcloud
- Install Cospend within Nextcloud
- Create a project in Cospend, add two members, then add 1 bill for member 1, then for the project click on the 3 dots in the left menu bar, click on "Reimbursement plan" (I don't know the exact english title) then on the Autosettlement button. The sums in the left menu, per user, should be updated. They are not (unless you disable the cache in the browser's devtools). (See https://github.com/eneiluj/cospend-nc/issues/134)

### Expected behavior

The JSON returned when calling https://mydomain.tld/apps/cospend/projects/courses should be from the server, not the browser's cache.

I expect the nextcloud_ynh app to disable cache for such requests.

nicofrand:
See https://github.com/nextcloud/documentation/issues/5472

nicofrand:
Add to nginx.conf file before `server {`:
```
# Expires mappings
map $sent_http_content_type $expires {
        default                                 off;
        text/html                               epoch;
        text/css                                max;
        application/javascript  max;
        ~image/                                 max;
}
```

then to nginx.conf replace `expires 1m;` with `expires $expires;`.

PR #486, opened 2022-05-02
Update nginx.conf

ericgaspar:
#484

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -1,3 +1,12 @@
+# Expires mappings
+map $sent_http_content_type $expires {
+        default                                 off;
+        text/html                               epoch;
+        text/css                                max;
+        application/javascript  max;
+        ~image/                                 max;
+}
+
 location ^~ /.well-known {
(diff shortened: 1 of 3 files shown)


ericgaspar:
!testme

conf/nginx.conf
```diff
@@ -1,3 +1,12 @@
+# Expires mappings
+map $sent_http_content_type $expires {
```
nicofrand, line 2: From my tests (I hardcoded it on my server) this need to be on the same level as `server {`. Did you test it?

ericgaspar:
!testme

conf/nginx.conf
ericgaspar: > Did you test it?
No, I havn't.

nicofrand: I double-checked: http://nginx.org/en/docs/http/ngx_http_map_module.html#map, `map` can only be defined in the `http` context, thus at the same level as the `server`.

PR #496, opened 2022-07-16
Enable-MySQL-4-byte

ericgaspar:
## Problem

- *Description of why you made this PR*

## Solution

- *And how do you fix that problem*

## PR Status

## Automatic tests

### DIFF
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -246,7 +246,7 @@ then
     exec_occ -V
 
     if [ "$(exec_occ config:system:get mysql.utf8mb4)" != "true" ]; then
-        exec_occ db:convert-mysql-charset
+        exec_occ maintenance:repair
     fi
 
     # Upgrade may fail if this app is enabled
--- a/scripts/upgrade
+++ b/scripts/upgrade
@@ -246,6 +246,8 @@ then
     exec_occ -V
 
     if [ "$(exec_occ config:system:get mysql.utf8mb4)" != "true" ]; then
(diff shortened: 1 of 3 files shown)


ericgaspar:
!testme

ericgaspar:
!testme

scripts/upgrade
```diff
@@ -246,7 +246,7 @@ then
     exec_occ -V
 
     if [ "$(exec_occ config:system:get mysql.utf8mb4)" != "true" ]; then
-        exec_occ db:convert-mysql-charset
+        exec_occ maintenance:repair
```
JimboJoe, line 249: Shouldn't you set also the `mysql.utf8mb4` attribute to true...?

ericgaspar: True...  I guess we also need to alter the database character set... (I am not able to test this on my setup for now...)

ericgaspar:
!testme

JimboJoe (approved):
Tested on my production server and that worked well (it takes a lot of time) 👍

JimboJoe:
From the (official doc)[https://docs.nextcloud.com/server/22/admin_manual/configuration_database/mysql_4byte_support.html], aren't we missing this part:
>Also make sure your backup strategy still work. If you use mysqldump make sure to add the --default-character-set=utf8mb4 option. Otherwise your backups are broken and restoring them will result in ? instead of the emojis, making files inaccessible.

I'm a bit puzzled because wallabag2 already manages utf8mb4 charset, but there's nothing specific in the package... 🤔

PR #530, opened 2022-12-09
Update to php8.1

ericgaspar:
- Update to PHP8.1 as recommanded in [25.0.2 release](https://nextcloud.com/fr/blog/maintenance-releases-25-0-2-24-0-8-and-23-0-12-are-out/)

PR #532, opened 2022-12-10
Testing

ericgaspar:
- Update to 25.0.3
- Only warn about 32-bit deprecation
- Update to recommanded PHP8.1


ericgaspar:
!testme

lapineige:
Pleas merge #533  first

tituspijean (approved):

lapineige:
Upgrade didn't work fine on my side as it triggered an internal server error, and the log still shows the same error :(

lapineige:
That's a quota issue, see: https://github.com/nextcloud/server/pull/34905#issuecomment-1345363531

lapineige:
Fixed by this manual change : https://github.com/nextcloud/server/pull/34905#issuecomment-1345388301

lapineige:
Let's merge ?

ericgaspar:
@lapineige It is not extra clear to me if user with 32 bits arch machines will still have to manually apply this tweaks https://github.com/nextcloud/server/pull/34905#issuecomment-1345388301

lapineige:
PR ongoing on Nextcloud side to fix that bug https://github.com/nextcloud/server/pull/35734.
Let's wait for it ?

Thatoo:
Raspi 4 are supposed to be arm64 not 32 bits right?
How is it then that I got the warning and could not update?
Will it be solved one this testing is merged into master?

lapineige:
> Raspberry pi 4 are supposed to be arm64 not 32 bits right?

Yes, but they used to be shipped with 32 bits OS by default.

ericgaspar:
Shall we merge this? This PR allows 32bit to upgrade Nextcloud

ericgaspar:
Few (maybe not enougth) 32bit use have tested this branch without this issue.

tituspijean (dismissed):

Gredin67:
25.0.3 was released some days ago,  including bugfixes. Wouldn't it be better to upgrade directly to it?

> PR ongoing on Nextcloud side to fix that bug [nextcloud/server#35734](https://github.com/nextcloud/server/pull/35734) (update: it need testers). Let's wait for it ?

Gredin67:
and also remove the warning about 32bit support deprecation, or at least amend it to follow this post https://forum.yunohost.org/t/nextcloud-may-support-32bits-for-above-version-26/23077

ericgaspar:
!testme

ericgaspar:
Clearly the trend is more like dropping 32 bit version. User are still reporting issues with 32 bits So I suggest we leave a warning about version 25 being the last to support 32 bit.

Gredin67:
If nextcloud actually continues an effort to maintain 32bits, why not follow them. Dropping nextcloud for 32bits will mean software obsolescence of many small servers, dropped to the trash.. 

So my opinion is we should not communicate warnings that are worse than what nextcloud actually plans. So something like
"Info: Nextcloud might drop support for 32bits in version 25, but probably more 26 or later, see https://forum.yunohost.org/t/nextcloud-may-support-32bits-for-above-version-26/23077"

ericgaspar:
Well yes, the idea is to keep 32-bit compatibility as much as possible. We also need to inform 32-bit user that support will soon be dropped.
How soon is soon? It doesn't seem clear upstream

lapineige:
> @Gredin67 If you have a better warning formulation please change it

`Nextcloud might drop 32bit support in the future. It is recommended that you upgrade to a 64 bits architecture.`

This would not add to specific detail about a version number that could fluctuate/be outdated very soon, not urge too much our users to upgrade, and still warn then that they should plan it somewhere in the quite near future.
What do you think ?

lapineige (dismissed):

lapineige:
https://github.com/nextcloud/server/pull/35734#issuecomment-1369025742
25.0.3 should be fixing the bug, at least most of the time, so I would vote for a merge. That would release the burden of many people not able to upgrade because of the previous (blocking) warning (and also lighten our support work), and for the few that might still encounter the bug there is a manual fix available.
Last CI and then we merge ?

lapineige:
!testme

lapineige:
Oh and let's merge #530 too ?

lapineige:
Upgrade went fine on my side.

scripts/upgrade
```diff
@@ -34,10 +34,10 @@ fpm_usage=$(ynh_app_setting_get --app=$app --key=fpm_usage)
 
 upgrade_type=$(ynh_check_app_version_changed)
 
-# Check machine architecture (Seens 25.0 Nextcloud doesn't support 32bit machines)
+# Check machine architecture (Nextcloud will soon deprecate 32-bit support)
 if [ $YNH_ARCH == "i386" ] || [ $YNH_ARCH == "armhf" ]
 then
-    ynh_die --message="Sorry, Nextcloud has deprecated 32-bit support"
+    ynh_print_warn --message="Nextcloud will soon deprecate 32-bit support. It is recommended to upgrade to a 64-bit architecture."
 fi
```
Gredin67, line 41: Nextcloud will deprecate 32-bit support within the next years. You should start thinking about upgrading to a 64-bit architecture.

lapineige:
!testme

lapineige:
Install failed for some strange reason :thinking:

ericgaspar:
This is just the CI failing a successfully installed app

lapineige:
CI is fully happy.

Let's merge ?

tituspijean (approved):
Tested and working!

PR #541, opened 2023-01-01
augment "opcache.revalidate_freq" from 1 to 30

OniriCorpe:
## Problem

currently the configuration files are checked almost every time nextcloud is loaded, which seems a bit excessive, especially on small installations with cheap hardware

> By default, cached scripts are revalidated on access to ensure that changes on disk take effect after at most 2 seconds. Since Nextcloud handles cache revalidation internally when required, the revalidation frequency can be reduced or completely disabled to enhance performance. Note, however, that it affects manual changes to scripts, including config.php. To check for changes at most every 60 seconds, use the following setting:
>
> `opcache.revalidate_freq = 60`

https://docs.nextcloud.com/server/25/admin_manual/installation/server_tuning.html#enable-php-opcache

## Solution

I increased the value to 30 seconds, as a hazardous compromise, but in my opinion it could be increased much more

## PR Status


OniriCorpe:
!testme

lapineige:
I would go for an higher value as except during upgrades (which reload them anyway) most of the time there is no reason these script would be changed.
But this as to come with an explanation (in the readme ?) that you need to reduce this value or rather reload Nextcloud (how ?) once you changed `config.php` and other scripts. If not, this will confuse many people and generate frustration + non-useful "bug reports".

PR #535, opened 2022-12-13
opcache.interned_strings_buffer higher 8

pp-r:
Because I got a warning about it in the Nextcloud settings. The recommendation should be higher than 8 so randomly i've put 10, restart php-fpm and no more warning.

## Problem

- *Description of why you made this PR*

## Solution

- *And how do you fix that problem*

## PR Status

- [ ] Code finished and ready to be reviewed/tested
- [ ] The fix/enhancement were manually tested (if applicable)

## Automatic tests

Automatic tests can be triggered on https://ci-apps-dev.yunohost.org/ *after creating the PR*, by commenting "!testme", "!gogogadgetoci" or "By the power of systemd, I invoke The Great App CI to test this Pull Request!". (N.B. : for this to work you need to be a member of the Yunohost-Apps organization)

PR #526, opened 2022-11-13
Fix upgrade of old versions <=v21 and for v22 to v23

SquirrelMe:
## Problem

While upgrading from nextcloud 19.0.3-ynh1 to the latest version 24.0.7-ynh1 on a yunohost 11.0.10.2 system I ran into two problems:

### Problem 1
The upgrade script identifies php8.x as the system's php version and tries to execute all incremental nextcloud updates with this version. Currently there is only a backward compatibility check that catches nextcloud versions from v14 downwards and executes them with php7.0. However, nextcloud seems to only support php8.x since v22.

Another user experienced the same problem here: https://github.com/YunoHost-Apps/nextcloud_ynh/issues/524, or, see my log https://paste.yunohost.org/raw/unupucanay at timestamp 2022-11-12 18:33:59,519.

### Problem 2
There seems to be a general nextcloud bug which affects certain upgrades from nextcloud 22 to 23. If the minor version of nextcloud 22 is too small (the yunohost nextcloud app upgrade script upgrades to X.0.0 versions, so 22.0.0) an upgrade is not possible. This is also reported in the nextcloud repository here: https://github.com/nextcloud/docker/issues/1652#issuecomment-986035522

Another user experienced the same problem here: https://github.com/YunoHost-Apps/nextcloud_ynh/issues/520 with corresponding log https://paste.yunohost.org/raw/qibogatayo at timestamp 2022-10-25 23:12:19,668, or, see my log https://paste.yunohost.org/raw/pogakopiyo at timestamp 2022-11-12 22:54:24,528.

### Solution 1
I extended the backward compatibility checks to use php7.4 for nextcloud versions 15 to 21.

### Solution 2
I increased the nextcloud version for v22 in the upgrade path to the latest 22.2.10 version which seems not to be affected by the upgrade bug.

## PR Status

- [x] Code finished and ready to be reviewed/tested
- [x] The fix/enhancement were manually tested (if applicable)

## Automatic tests

Automatic tests can be triggered on https://ci-apps-dev.yunohost.org/ *after creating the PR*, by commenting "!testme", "!gogogadgetoci" or "By the power of systemd, I invoke The Great App CI to test this Pull Request!". (N.B. : for this to work you need to be a member of the Yunohost-Apps organization)

PR #555, opened 2023-02-03
Testing

ericgaspar:
- Set `opcache.interned_strings_buffer` to 16 #535
-  Fix upgrade of old versions <=v21 and for v22 to v23 #526


ericgaspar:
!testme

ericgaspar:
!testme

PR #574, opened 2023-06-02
Fix /ocm-provider/” and “/ocs-provider/” on NGINX server warnings

ericgaspar:
fix for [“Not properly set up to resolve “/ocm-provider/” and “/ocs-provider/” on NGINX server](https://help.nextcloud.com/t/with-nextcloud-update-not-properly-set-up-to-resolve-ocm-provider-and-ocs-provider-on-nginx-server/158528) warnings


ericgaspar:
!testme

PR #592, opened 2023-07-28
26.0.11 (Oldstable)

zamentur:
## Problem

Need a branch of the repo with 26.0.11 for people who wants to stay a bit on 26 branch...

## Solution

Don't merge or delete this branch, i suggest to use it as a main branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ In addition to Nextcloud core features, the following are made available with th
  * Serve `/.well-known` paths for CalDAV and CardDAV on the domain only if it's not already served - i.e. by Baïkal
 
 
-**Shipped version:** 28.0.1~ynh1
+**Shipped version:** 28.0.1~ynh2
 
 **Demo:** https://demo.nextcloud.com/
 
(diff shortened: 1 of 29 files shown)


zamentur:
!testme

lapineige:
It might be very important for the people locked on 32-bits architecture.

alexAubin:
(Do you mean 26.x is the last supported version for 32 bit arch ? I thought Nextcloud somehow reconsidered their decisions and were not going to drop 32 bit support before a while ? Or maybe I missed some new part of the story :grimacing: )

lapineige:
That is right : they announced they will try to continue to support 32 bit.
But we don't know for how long, and in v26 (should be fixed in the last minor releases) and v27 there were regression specific to the 32 bit architecture, that took a while to be fixed.
So having the hotfix backported to v26 in Yunohost is great for such situations :)

tomdereub:
Hi, it would be grate to get 26.0.6 on this branch, if someone could do that.

ericgaspar:
!testme

tomdereub:
Hi, thanks a lot, is it already ready to upgrade ?

ericgaspar:
!testme

zamentur:
!testme

ericgaspar:
!testme

ericgaspar:
!testme

PR #636, opened 2023-12-04
[fix] redirect logout

zamentur:
## Problem

Logout on sso was broken and don't return on the nextcloud login form.

## Solution

Fix all the things and logout on sso only if it's a ldap user (it's possible to create internal nextcloud user side to the ldap users...)

## PR Status

## Automatic tests

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -1,33 +1,11 @@
 #!/bin/bash
 
-#=================================================
-# GENERIC START
-#=================================================
-# IMPORT GENERIC HELPERS
-#=================================================
-
 source _common.sh
 source /usr/share/yunohost/helpers
 source _ynh_mysql_connect_as.sh
 
-#=================================================
-# STORE SETTINGS FROM MANIFEST
-#=================================================
-
-maintenance_mode=0
-fpm_footprint="high"
-fpm_free_footprint=0
-fpm_usage="medium"
 phpflags="--define apc.enable_cli=1"
 
-#=================================================
-# STORE SETTINGS FROM MANIFEST
-#=================================================
-
-ynh_app_setting_set --app=$app --key=maintenance_mode --value=$maintenance_mode
-ynh_app_setting_set --app=$app --key=fpm_footprint --value=$fpm_footprint
-ynh_app_setting_set --app=$app --key=fpm_free_footprint --value=$fpm_free_footprint
-ynh_app_setting_set --app=$app --key=fpm_usage --value=$fpm_usage
 ynh_app_setting_set --app=$app --key=phpflags --value=$phpflags
 
 #=================================================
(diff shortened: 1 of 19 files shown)


ericgaspar:
!testme

ericgaspar:
Any way to test this? I have upgraded my server with this PR but can't find any noticable difference.

zamentur:
> Any way to test this? I have upgraded my server with this PR but can't find any noticable difference.

What do you mean exactly ?

With this PR, when i click on nextcloud logout: 

- If i am a LDAP user
  -  I am logout from YunoHost SSO
  -  I am logout from nextcloud
  -  I am redirected on nextcloud forms
- If i am a nextcloud internal user:
  -  I am logout from nextcloud
  -  I am redirected on nextcloud forms

Without this PR, but with the Testing PR instead: 
 - the logout patch failed (see https://github.com/YunoHost/issues/issues/2296 )
 - logout from nextcloud doesn't work at all, if you are logged in with the SSO...

ericgaspar:
During my test, when connected to Nextcloud with LDAP user, when I disconnect from Nextcloud there is no effect, I stay on Nextcloud page. The patch should be reapplied during upgrade right?

zamentur:
> I stay on Nextcloud page

without being logged out ?

ericgaspar:
> > I stay on Nextcloud page
> 
> without being logged out ?
Yes, without being logged out.

https://github.com/YunoHost-Apps/nextcloud_ynh/assets/46165813/d18ae263-17fd-4589-949d-f256a4095c52

zamentur:
Now it works if we are in UPGRADE_APP mode, but we probably should be better.

ericgaspar:
I have upgraded with latest changes but still have the same behaviors as before.
What do you mean by UPGRADE_APP mode?

zamentur:
UPGRADE_APP mode is triggered if the sources change or if we run the upgrade with `--force`.

zamentur:
https://github.com/YunoHost/yunohost/blob/dev/src/app.py#L669

ericgaspar:
@zamentur Is it working on your side?

zamentur:
- [x] Fresh install with subpath: yes
- [ ] Upgrade with subpasth: no (The upgrade work only if you upgrade from a smaller upstream version :/ )

zamentur:
Note that 27.1.5rc1 is on the road...

ericgaspar:
strangely #644 PR has solved my redirect logout issue

zamentur:
I know we have a oldstable branch, however, i think we probably should avoid to upgrade on recent new major version, there are always bugs with X.0.x version...

lapineige:
In that case we are already in version 28.0.1

ericgaspar:
!testme

scripts/install
```diff
@@ -164,12 +164,13 @@ fi
...
 
 # Add dynamic logout URL to the config
+url_base64="$(echo -n "https://$domain$path" | base64)"
 exec_occ config:system:get logout_url >/dev/null 2>&1 \
   || echo "
 //-YunoHost-
 // set logout_url according to main domain
 \$main_domain = exec('cat /etc/yunohost/current_host');
-\$CONFIG['logout_url'] = 'https://'.\$main_domain.'/yunohost/sso/?action=logout';
+\$CONFIG['logout_url'] = 'https://'.\$main_domain.'/yunohost/sso/?action=logout&r=${url_base64}';
```
alexAubin, line 173: I'm not sure to understand why the main domain is used here compared to just `$domain` on which the app is installed ...

Anyway, this will probably stop working on 12.x / Bookworm because portals will now be "independent" in terms of "main domain trees", so if Nextcloud is installed on a different "main domain" this ain't gonna work

Also the `/yunohost/sso/?action=logout` endpoint is probably going to be `/yunohost/sso/logout`

But anyway there's no way to make this future-proof i think, we'll just have to patch Nextcloud once we'll be here

PR #637, opened 2023-12-08
[enh] Faster permissions setter on big datadir

zamentur:
## Problem

With a lot of data 100GB+ , nextcloud took too long to upgrade.

## Solution

I believe the solution in this stackoverflow is faster https://stackoverflow.com/a/41331598

## PR Status

## Automatic tests


zamentur:
On my test, for 160GB the old code took 28s and the new one 7s.

I was thinking it was in minutes for the old one (i observed minutes during a true upgrade of nextcloud...), so i don't know what to say.

ericgaspar:
!testme

scripts/install
```diff
@@ -259,10 +259,8 @@ ynh_multimedia_addaccess $app
 # Fix app ownerships & permissions
 chown -R $app:www-data "$install_dir"
 chown -R $app: "$data_dir"
-find $install_dir/ -type f -print0 | xargs -r0 chmod 0644
-find $install_dir/ -type d -print0 | xargs -r0 chmod 0755
-find $data_dir/data/ -type f -print0 | xargs -r0 chmod 0640
-find $data_dir/data/ -type d -print0 | xargs -r0 chmod 0750
+chmod -R a=r,u+w,a+X $install_dir/
```
alexAubin, line 262: (Any reason we have to have `755` for the install_dir rather than `750` ? x_X)

scripts/install
```diff
@@ -259,10 +259,8 @@ ynh_multimedia_addaccess $app
...
+chmod -R ug=r,o=---,u+w,ug+X $data_dir/data/
```
alexAubin, line 263: It feels pretty convoluted to have `ug=r` followed by `u+w` and `ug+X`, we should be able to write directly:

```suggestion
chmod -R u=rwX,g=rX,o=--- $data_dir/data/
```

zamentur: Sadly, it's not equivalent: with your notation, if a file has a x permission for user, this permission x is not removed (yes i know X behavior is very crazy), with mine the x permission is removed cause we first set u=r...

But your comments let's say that adding a comment on that could be a good idea.

scripts/install
zamentur: I don't know, but some lines after we put 750 on the dir (without recursive option) so i guess those files are protected anyway from others users...

ericgaspar:
!testme

scripts/upgrade
```diff
@@ -358,10 +358,8 @@ exec_occ background:cron
 # Fix app ownerships & permissions
 chown -R $app:www-data "$install_dir"
 chown -R $app: "$data_dir"
-find $install_dir/ -type f -print0 | xargs -r0 chmod 0644
-find $install_dir/ -type d -print0 | xargs -r0 chmod 0755
-find $data_dir/data/ -type f -print0 | xargs -r0 chmod 0640
-find $data_dir/data/ -type d -print0 | xargs -r0 chmod 0750
+chmod -R a=r,u+w,a+X $final_path/
```
Thovi98, line 361:
```suggestion
chmod -R a=r,u+w,a+X $install_dir/
```

alexAubin:
!testme

scripts/install
```diff
@@ -259,10 +259,8 @@ ynh_multimedia_addaccess $app
 # Fix app ownerships & permissions
 chown -R $app:www-data "$install_dir"
 chown -R $app: "$data_dir"
```
kay0u, line 261: On my server, the main issue is this line. Idk how to speed up this chown though

alexAubin: Quick look on the internet : https://unix.stackexchange.com/questions/245274/chown-only-where-needed-speedup-chown

Apparently using `find` to only apply `chown` for the relevant file is way faster, could be nice to give it a try

zamentur: If you have a lot of files, the find method could trigger a message saying there is too many destination listed (if i remeber an old bug)

Thovi98: It seems `chown -R $app: $data_dir` is performed 3 times during upgrade, which shouldn’t be necessary because `$data_dir` is not touched during the upgrade as it is in another place than `$install_dir` ? (cf. https://docs.nextcloud.com/server/latest/admin_manual/maintenance/manual_upgrade.html)

kay0u:
closed by #711

PR #644, opened 2023-12-21
Fix patches for v28

lapineige:
## Problem

- https://github.com/nextcloud/server/blame/master/core/Controller/LoginController.php#L102

### DIFF
--- a/sources/patches_last_version/app-00-add-logout_url-conf.patch
+++ b/sources/patches_last_version/app-00-add-logout_url-conf.patch
@@ -8,11 +8,11 @@ index 13aef8f67a..55d8dbf9d1 100644
  
 -		$response = new RedirectResponse($this->urlGenerator->linkToRouteAbsolute(
 -			'core.login.showLoginForm',
--			['clear' => true] // this param the the code in login.js may be removed when the "Clear-Site-Data" is working in the browsers
+-			['clear' => true] // this param the code in login.js may be removed when the "Clear-Site-Data" is working in the browsers
 +		$redirectUrl = $this->config->getSystemValue('logout_url',
 +			$this->urlGenerator->linkToRouteAbsolute(
 +				'core.login.showLoginForm',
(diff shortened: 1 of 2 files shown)


lapineige:
!testme

lapineige:
!testme

lapineige:
Fix patch fixes the issue with LoginController.php

```
153105 INFO DEBUG - patching file core/Controller/LoginController.php
153105 INFO DEBUG - Hunk YunoHost-Apps#1 succeeded at 97 (offset -22 lines).
```

From https://ci-apps-dev.yunohost.org/ci/job/12038

alexAubin:
(NB : randomly jumping in, but @zamentur was pointing the other day that it is less trivial than it looks, because during the upgrade from older major version, the current upgrade process tries to applies the "new" patch on old versions, which is likely to fail ... Somehow we should either *not* apply the patch until the latest version, or apply the old version of the patch ...)

lapineige:
Uh, I already thought it was trivial to fix :laughing:

This is in the last-version folder, maybe it already handled ?

lapineige:
> Alrighty! [Test Badge](https://ci-apps-dev.yunohost.org/ci/job/12040) [](https://ci-apps-dev.yunohost.org/ci/job/12040)

Here we go, patches fixed for install :tada:

alexAubin:
Hmmyeah my bad it seems to already be handled since a long time :P https://github.com/YunoHost-Apps/nextcloud_ynh/blob/83ac399638fe44f7a50e9de51d4f8cdb4659333b/scripts/upgrade#L197

lapineige:
Upgrades are working straight away in CI :tada:

I'll try manual upgrade from v27.

lapineige:
Should we update test.toml to check also upgrade from v27 ?

lapineige:
Change_url fails too but it seems to be a CI issue.

ericgaspar (approved):

Issue #650, opened 2024-01-15
UPGRADE 22->28 : syntax error in control file: 8.x, php7.4


samuelvermeulenpro:
Hello All and Good Year,

### Describe the bug

*syntax error in control file: 8.1(and 8.2), php7.4*

### Context

- Hardware: *RaspberryPI 5*
- YunoHost version: 11.2.x
- I have access to my server: *Through SSH*
- Are you in a special context or did you perform some particular tweaking on your YunoHost instance?: *no*

### Steps to reproduce

- *command from the CLI, :*
    ```yunohost app upgrade nextcloud -u https://github.com/YunoHost-Apps/nextcloud_ynh --debug ```

#### Logs

```
...
2024-01-15 16:06:01,666: DEBUG - syntax error in control file: 8.1, php7.4
2024-01-15 16:06:01,668: DEBUG - + LC_ALL=C
2024-01-15 16:06:01,668: DEBUG - + dpkg --force-depends --install ./nextcloud-ynh-deps_28.0.1~ynh1_all.deb
2024-01-15 16:06:01,669: DEBUG - + tee ./dpkg_log
2024-01-15 16:06:01,684: DEBUG - dpkg: error: cannot access archive './nextcloud-ynh-deps_28.0.1~ynh1_all.deb': No such file or directory
...
2024-01-15 16:06:04,453: DEBUG - + phpversion='7.4;8.1'
2024-01-15 16:06:04,453: DEBUG - + [[ 8.2 != \7\.\4\;\8\.\1 ]]
...
```
Complete logs : https://paste.yunohost.org/raw/acumuroyag

Why `7.4;8.1` ? And after auto-restoring the nextcloud old version, into `/etc/php` directory there is an illegal named folder `7.4;8.1` !! Bug or not bug ?

Thank you for your suggestions ;)

alexAubin:
technical note : this is most likely due to https://github.com/YunoHost/yunohost/blob/dev/helpers/apt#L295 which we should reset in `exec_occ` - the issue occurs when the upgrade goes through several PHP versions

alexAubin:
(Fixed in latest release)

PR #652, opened 2024-01-21
Fix PHP/APT dependency confusion when upgrading through many versions

alexAubin:
## Problem

- Fix https://github.com/YunoHost-Apps/nextcloud_ynh/issues/650 , with the code ending up with `phpversion='7.4;8.1'` because `ynh_install_app_dependencies` is called several times, and the second time it *appends* (rather than *replace*) the list of dependencies

## Solution

- Force the helper to replace the dependencies


alexAubin:
!testme

PR #653, opened 2024-01-21
New source format

alexAubin:
## Problem

- We have to use ugly tricks to create old `app.src` on the fly when we could use the new manifest format
- But we have a shitload versions in `upgrade.d/` and who really believes that upgrading from Nextcloud 15.x still works x_x
- Apparently upgrading from 22.0 did not work because Nextcloud wants to be on 22.2 before upgrading to 23.x

## Solution

- Rework source management using the new source syntax in manifest.toml
- Drop support for versions before 22.2 with a pointer to the appropriate command for folks willing to upgrade to 22.2 first

### DIFF
--- a/doc/ADMIN.md
+++ b/doc/ADMIN.md
@@ -3,7 +3,7 @@
 You can run Nextcloud commands from the command line using:
 
 ```
-sudo -u __APP__ php__YNH_PHP_VERSION__ --define apc.enable_cli=1 __INSTALL_DIR__/occ ...
+sudo -u __APP__ php__PHPVERSION__ --define apc.enable_cli=1 __INSTALL_DIR__/occ ...
 ```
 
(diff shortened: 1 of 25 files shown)


alexAubin:
!testme

lapineige:
Yes at some point we have to drop supporting automated update from years old versions… It's just too much work.

[Nextcloud 22 is from July 2021](https://nextcloud.com/blog/nextcloud-hub-22-introduces-approval-workflows-integrated-knowledge-management-and-decentralized-group-administration/).
2.5 years and 6 versions.

That's more than enough in my opinion.

Users who still did not updated their instance since almost 3 years (which might happens if you restore an old backup for instance) will still be able to do it, by upgrading manually to a specific commit (refering to that version) in the repository history.


Thanks you for that hard, painful work :clap:

alexAubin:
(Imho this is done can be merged to testing)

PR #661, opened 2024-01-31
Modify HSTS Preload policy duration to 1 year

frju365:
## Problem

*HSTS policy duration is too short* for these reasons : 
- For now the duration of the HSTS policy is set to 6 months which is the minimum duration. It's recommended for more security to have at least 1 year. cf. https://infosec.mozilla.org/guidelines/web_security#http-strict-transport-security
- To add the domain to the HSTS preload list (which is recommended to preload the domain directly in the browser), the duration must be > 1 year cf. https://hstspreload.org/#deployment-recommendations

Sources : 
* https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Strict_Transport_Security_Cheat_Sheet.html
* https://datatracker.ietf.org/doc/html/rfc6797

## Solution

- *Set the HSTS policy duration to 1 year*

## PR Status

## Automatic tests

### DIFF
--- a/conf/nginx.conf
+++ b/conf/nginx.conf
@@ -55,7 +55,7 @@ location ^~ __PATH__/ {
   client_body_buffer_size 512k;
 
   # HTTP response headers borrowed from Nextcloud `.htaccess`
-  more_set_headers "Strict-Transport-Security: max-age=63072000; includeSubDomains; preload;";
+  more_set_headers "Strict-Transport-Security: max-age=31536000; includeSubDomains; preload;";
   more_set_headers "Referrer-Policy: no-referrer";
   more_set_headers "X-Content-Type-Options: nosniff";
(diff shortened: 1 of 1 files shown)


frju365:
!testme

alexAubin (approved):
Why not but I'm confused why we have these headers in the Nextcloud app specifically, yet the HSTS is already set by the core https://github.com/YunoHost/yunohost/blob/dev/conf/nginx/server.tpl.conf#L111, and similar stuff for the other headers such as Download-Options etc...

frju365:
Hello Alex ! Don't know too. It would be better indeed to do modification in global template.

ericgaspar:
closing with [43c5bab](https://github.com/YunoHost-Apps/nextcloud_ynh/pull/662/commits/43c5bab777ce7ca455a212635a3c800748b6ea84)

PR #638, opened 2023-12-10
Nextcloud parameters in config panel

zamentur:
## Problem

Some configuration parameters need to be adapt according to localization, kind of network, security models, ldap or internal nextcloud account, disk space available, cpu, etc.

## Solution

Add some parameters to the config panel

## PR Status

- [ ] Code finished and ready to be reviewed/tested
- [ ] The fix/enhancement were manually tested (if applicable)

## Automatic tests

Automatic tests can be triggered on https://ci-apps-dev.yunohost.org/ *after creating the PR*, by commenting "!testme", "!gogogadgetoci" or "By the power of systemd, I invoke The Great App CI to test this Pull Request!". (N.B. : for this to work you need to be a member of the Yunohost-Apps organization)

PR #664, opened 2024-02-09
Add language choice for installation

CodeShakingSheep:
## Problem

Nextcloud shows the following warning:

Your installation has no default phone region set. This is required to validate phone numbers in the profile settings without a country code. To allow numbers without a country code, please add "default_phone_region" with the respective ISO 3166-1 code of the region to your config file. For more details see the [documentation ↗](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements).

image

## Solution

Added question for language in `manifest.toml`. Set 3 language variables (see here: https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#default-language ) in `install`:
- `default_language`
- `default_locale`
- `default_phone_region`

~~In `upgrade` I set `default_phone_region` if it's not set according to `default_locale` (which always has a value as it has a default value).~~

Edit: Setting `default_phone_region` for existing installations should be handled in config panel, cf. #638. 

## PR Status

## Automatic tests


CodeShakingSheep:
!testme

CodeShakingSheep:
!testme

CodeShakingSheep:
!testme

scripts/upgrade
```diff
@@ -143,6 +143,10 @@ then
     # Fix warning "Your installation has no default phone region set"
     if [ "$(exec_occ config:system:get default_phone_region)" == "" ]; then
 	config_locale=$(exec_occ config:system:get default_locale)
+	if [ "$config_locale" == "" ]; then
+		# If it's not set in config use "US" as default
+		config_locale="US"
```
ericgaspar: Why not FR by default?

jarod5001: Why not any other code. It would be more logical that the user chooses from a drop down list at the installation or change it in the app panel

ericgaspar: We are talking about a default value set if there was no previous value set.

CodeShakingSheep: Hi @ericgaspar,
I set "US" because in `manifest.toml` I used "en_US" as default install language and so I wanted to use the default country here as well.

CodeShakingSheep:
!testme

CodeShakingSheep:
> Alrighty! [Test Badge](https://ci-apps-dev.yunohost.org/ci/job/13727) [](https://ci-apps-dev.yunohost.org/ci/job/13727)

What is going on on the test machine? There is this error which is completely unrelated to this package. Can someone look into that?

```
WARNING W: GPG error: https://packages.sury.org/php bullseye InRelease: The following signatures were invalid: EXPKEYSIG B188E2B695BD4743 DEB.SURY.ORG Automatic Signing Key <deb@sury.org>
```

scripts/install
```diff
@@ -87,6 +87,17 @@ ynh_script_progression --message="Configuring $app..." --weight=8
 # Set the mysql.utf8mb4 config to true in config.php
 exec_occ config:system:set mysql.utf8mb4 --type boolean --value="true"
 
+# Set default language in config.php
+raw_language=$(echo "$language" | awk -F'_' '{print $1}')
+exec_occ config:system:set default_language --value="$raw_language"
```
Salamandar, line 92: shouldn't it be just $language here ?

scripts/upgrade
ericgaspar: let's then change the default language to `fr` (because it is by far the most used...)

scripts/install
CodeShakingSheep: Nope, Nextcloud's `default_language` uses ISO_639-1 language codes such as `en` for English, see https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#default-language

That's why we need to cut the first part from `$language`, e.g. take `en` from `en_US`.

Tagadda:
I restarted the last CI job https://ci-apps-dev.yunohost.org/ci/job/13727

CodeShakingSheep:
> is this language choice not best handled with config panels? cf. #638

Thanks for the hint. I hadn't checked out that PR. Indeed it seems to be the cleaner solution to set a value in config panel rather than to use a fixed default value during upgrade. So, I removed the fix in `upgrade` and renamed this PR accordingly.

CodeShakingSheep:
!testme

alexAubin:
Closing because it's been almost one year and it should really be something in the config panel rather than in the install questions ... in fact it's handled in https://github.com/YunoHost-Apps/nextcloud_ynh/pull/638 which does a bunch of other things, so closing in favor or #638

PR #671, opened 2024-02-29
fix pre upgrade message

OniriCorpe:
## Problem

- the actual pre upgrade message is showed in any case, however it is not relevant if already in version 28

## Solution

- move it to `doc/PRE_UPGRADE.d` and rename it `28.0.3~ynh1.md`

## PR Status


alexAubin:
Eeeeh yeah but the thing is that we'll want this for every major version ... I guess we can duplicate it once 29.x, 30.x, etc will be out, but people jumping multiple major version will have it displayed multiple times ... But yeah the link to the forum topic is related to 28.x so meh idk watdo

OniriCorpe:
there is not one major version a week, it's fine i guess x)

alexAubin:
Actually nevermind, once 29.x is out we can just edit+rename the PRE_UPGRADE file such that it's only shown when upgrading to the new major version so it's fine

alexAubin (approved):

PR #684, opened 2024-04-23
29.0.0

ericgaspar:
- Upgrade to 29.0.0

PR #688, opened 2024-05-23
Testing

ericgaspar:
- #684 29.0.2
- #690
- #691


ericgaspar:
!testme

nicofrand:
Hi!

I don't know if this is related to testing or also the case with master but I get this error in /admin/overview:

> Votre dossier de données et vos fichiers sont probablement accessibles depuis internet. Le fichier .htaccess ne fonctionne pas. Nous vous recommandons vivement de configurer votre serveur web de façon à ce que ce dossier de données ne soit plus accessible, ou de le déplacer hors de la racine du serveur web.

It basically says that the .htaccess is not working and my data is reachable through internet.

nicofrand:
:facepalm: That might only be a warning for apache and should be ignored? I don't know if nginx is secured enough though?

ericgaspar:
We have some rules borrowed from `.htaccess` in our NGINX config

kay0u (approved):

nicofrand:
> We have some rules borrowed from `.htaccess` in our NGINX config


Do you know if nextcloud really checks the file access or just the htaccess by another way?
Maybe the nginx config is broken too?

PR #690, opened 2024-06-01
Fix space splitting in phpflags on install

lyynd:
## Problem

`phpflags` ends up being `--define;apc.enable_cli=1` because of space splitting. I did not investigate why it breaks in this exact way, something inside `ynh_app_setting_set`. Because of this `php occ` in `yunohost app shell` does not work.

## Solution

- Use quotes to prevent space splitting.

## PR Status

## Automatic tests

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -6,7 +6,7 @@ source _ynh_mysql_connect_as.sh
 
 phpflags="--define apc.enable_cli=1"
 
-ynh_app_setting_set --app=$app --key=phpflags --value=$phpflags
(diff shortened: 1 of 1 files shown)


lyynd:
!testme

ericgaspar (approved):

lyynd:
@ericgaspar I am not sure if this is right place to discuss this, but we might also want to fix existing instances, since the wrong setting is set on install. This would require some kind of migration. What do you think?

PR #726, opened 2024-09-03
handle mimetype migrations

ericgaspar:
Fixes : One or more mimetype migrations are available. Occasionally new mimetypes are added to better handle certain file types. Migrating the mimetypes take a long time on larger instances so this is not done automatically during upgrades. Use the command `occ maintenance:repair --include-expensive` to perform the migrations.


ericgaspar:
!testme

Issue #612, opened 2023-10-12
Add trashbin cleanup and file versioning


tio-trom:
```
  'trashbin_retention_obligation' => 'auto, 30',
  'versions_retention_obligation' => 'auto, 30',
```

To the config file. This way we make sure that after 30 days the trashbin gets cleaned up. Also only keep file versioning for 30 days. I think it makes sense, else the server will fill up with trashed files.

Bugsbane:
Clearing the trashbin certainly makes sense as the user has specifically chosen to delete items and then not restored them for a month. I would be hesitant to set versions to 30 days though, as this could be an unexpected behavior to users leading to data loss.

We don't want old versions filling up the server though, but I would suggest that this would be better served by setting it to *not* delete automatically by default, but suggesting this config setting in the Yunohost app description and app docs so server admins will be aware if it is set.

tio-trom:
The file versioning means to only keep 30 versions of the same file. I think that's also reasonable.

Bugsbane:
Ah, ok. My mistake. I didn't read closely enough assumed it was also 30 days. In that case yes, both make good sense.

Tagadda:
If I understand correctly, `30, auto` is the same as `auto`

> `auto`: default setting. keeps files and folders in the trash bin for 30 days and automatically deletes anytime after that if space is needed (note: files may not be deleted if space is not needed).
> 
> `D, auto`: keeps files and folders in the trash bin for D+ days, delete anytime if space needed (note: files may not be deleted if space is not needed)

https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/trashbin_configuration.html

Tagadda:
I don't think the default behavior is good for most use cases.
We could handle file retention through the config panel tho'

lapineige:
Yeah as user won't be able to notice the change that easily, I would be more in favour of a min 30 days of retention + delete only if space needed, which serves the purpose of freeing space when it grows too much, but keep a safe month of trashbin history.

Tagadda:
Make admins notice the change easily, that's a mission for POST_UPGRADE.md o/

tootbrute:
For those of you who found this issue because your hard drive if full...

More info about the [trashbin options](https://docs.nextcloud.com/server/18/admin_manual/configuration_server/config_sample_php_parameters.html#deleted-items-trash-bin)

You can edit your config.php file here:
`/var/www/nextcloud/config/config.php`

[...]

PR #737, opened 2024-10-04
Doc tomdereub

tomdereub:
## Problem/Solution

Add doc about file retention, linked to https://github.com/YunoHost-Apps/nextcloud_ynh/issues/612.

## PR Status

- [x] Code finished and ready to be reviewed


MCMic:
The link to Nextcloud documentation should be to latest, not 18. Or to the packaged version in this repo, so hardcoded 29 or 30.

ericgaspar:
I feel that, even if the information is relevant in itself, we should not clutter our doc by duplicating that of Nextcloud.

tomdereub:
I would see 2 ways of looking at it : 
- yunohost package ships default nextcloud config, as for now, and doesn't "duplicate" nextcloud's doc.
- yunohost package includes some specific config, as demanded in https://github.com/YunoHost-Apps/nextcloud_ynh/issues/612, and the package doc must explain this

Yunohost goal is to ease server admin for people that have limited skills in server administration, so in my opinion it's better to add a specific config that will avoid problems for most users (full storage in that case), and document it.
But that's just my humble opinion...

PR #767, opened 2024-12-20
Fix upgrade exploding when $app is nextcloud__2, __3 etc because the virtual apt/dpkg package is named with - and not _

rodinux:
## Problem

- *when multi instance the $app-ynh-deps need be formatted with Replace all '_' by '-', and append -ynh-deps*

## Solution

- *correct syntax from helpers*

## PR Status

## Automatic tests


rodinux:
!testme

PR #774, opened 2025-01-04
Update post_user_delete

### LINKED ISSUE #773: post_user_delete hook throws error and does not run
### Describe the bug

When i try and delete a user, I get the following error:
```
Warning: ./50-nextcloud: line 13: ynh_exec_as_app: command not found
Error: Could not run script: /etc/yunohost/hooks.d/post_user_delete/50-nextcloud
```
The user is then unable to log in to nextcloud but they are not fully deleted.

Looking at the logs, the problem seems to be that the hook is trying to use app helpers version 1, where ynh_exec_as_app is (probably) not present

### Context

- YunoHost version: 12.0.9.1

### Steps to reproduce

- *If you performed a command from the CLI, the command itself is enough. For example:*
    ```sh
    sudo yunohost user delete <user>
    ```

### Expected behavior

The post_user_delete hook completes and the user is deleted from the nextcloud instance 

### Logs

https://paste.yunohost.org/raw/gevayugaqi

ericgaspar:
Fix #773

### DIFF
--- a/hooks/post_user_delete
+++ b/hooks/post_user_delete
@@ -1,16 +1,11 @@
 #!/bin/bash
 
+source ../settings/scripts/_common.sh
 source /usr/share/yunohost/helpers
 
 user="$1"
 app="$(basename $0 | cut -d- -f 2-)" # Extract the app name from the script name, which is supposed to be something like "50-app_id"
 install_dir=$(ynh_app_setting_get --app=$app --key=install_dir)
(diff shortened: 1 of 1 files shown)


ericgaspar:
!testme

Tagadda:
You won't be able to source the helpers like that.
Hooks are copied to `/etc/yunohost/hooks.d/post_user_create` during app install.

PR #779, opened 2025-01-30
Migrating the mimetypes

ericgaspar:
Occasionally new mimetypes are added to better handle certain file types.
Migrating the mimetypes take a long time on larger instances so this is not done automatically during upgrades.


ericgaspar:
!testme

PR #805, opened 2025-05-04
add versions_ and trashbin_retention_obligation

Gredin67:
## Problem

- #612 

## Solution

- [ ] add settings trashbin_retention_obligation and versions_retention_obligation

- [ ] add settings to config panel

- [ ] 

- see also https://docs.nextcloud.com/server/latest/user_manual/en/files/deleted_file_management.html

## PR Status

- [ ] Code finished and ready to be reviewed/tested
- [ ] The fix/enhancement were manually tested (if applicable)

## Automatic tests

Automatic tests can be triggered on https://ci-apps-dev.yunohost.org/ *after creating the PR*, by commenting "!testme", "!gogogadgetoci" or "By the power of systemd, I invoke The Great App CI to test this Pull Request!". (N.B. : for this to work you need to be a member of the Yunohost-Apps organization)

PR #806, opened 2025-05-05
Testing

ericgaspar:
- #805
- [31.0.5](https://github.com/YunoHost-Apps/nextcloud_ynh/pull/806/commits/6f77fc45d34340a8c378e3c143bd154eb22dd7b9)


Gredin67:
following discussions in #612 I was thinking of adding a setting in the config panel or at least warn the user that with this upgrade: 

- only up to 30 versions of a file will be kept 
- trashbin retention will be set to 30 days

ericgaspar:
!testme

PR #714, opened 2024-08-01
Manage oldstable version

kay0u:
## Problem

- *Description of why you made this PR*

## Solution

- *And how do you fix that problem*

## PR Status

- [ ] Code finished and ready to be reviewed/tested
- [ ] The fix/enhancement were manually tested (if applicable)

## Automatic tests

Automatic tests can be triggered on https://ci-apps-dev.yunohost.org/ *after creating the PR*, by commenting "!testme", "!gogogadgetoci" or "By the power of systemd, I invoke The Great App CI to test this Pull Request!". (N.B. : for this to work you need to be a member of the Yunohost-Apps organization)

PR #816, opened 2025-06-17
Manage oldstable version (#714)

zamentur:
## Problem
I squash merged https://github.com/YunoHost-Apps/nextcloud_ynh/pull/714 into this branch cause there are a lot of unrelated commits in history...

We need a way to follow the N-1 version (oldstable) cause it's generally, more stable, than stable :/

## Solution

- *And how do you fix that problem*

## PR Status

## Automatic tests

### DIFF
--- a/config_panel.toml
+++ b/config_panel.toml
@@ -25,6 +25,19 @@ name.fr = "Configuration de Nextcloud"
         help.en = "This action sets permissions for all data. Beware, this action can take up to several hours if users have a lot of data."
         help.fr = "Cette action définit les autorisations pour toutes les données. Attention, cette action peut prendre jusqu'à plusieurs heures si les utilisateurs ont beaucoup de données."
 
+    [main.version]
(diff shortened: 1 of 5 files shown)


alexAubin:
>We need a way to follow the N-1 version (oldstable) cause it's generally, more stable, than stable :/

Imho one option on the table would be just to invert things with : oldstable/N-1 being the version we ship by default, keep the "N" version in some sort of "next" version (or "testing" but that makes things a bit trickier when we want to hotfix the default branch), and then have a disclaimer somewhere explaining that the default version is the N-1 because it's considered to be more stable than version N, with a pointer on how to upgrade the the "next" version branch for people that absolutely want to upgrade early ...

tomdereub:
I think version names should be consistent with nextcloud's names. So keep calling stable for N and oldstable for N-1.
But I agree with shipping oldstable by default, with a short text that explains that it's more stable than stable.

PR #844, opened 2025-12-22
Testing / 32.0

ericgaspar:
- 32.0.3


luc38090:
VPS test server at OVH (wich is a daily Nextcloud Yunohost backup of my production server), with about 20Gb of files and a database of about 1,3Go
-> Upgrade from Nextcloud 31.0.11 to 32.0.3 without any issue (Map app has to be disabled because there is no release for Nextcloud 32)

- Files app: all the files and photos seems to be here
- The 4.0 Ubuntu client app for syncing files is able to do the job
- Agenda app: all the events seems to be here
- Contacts app: all contacts seems to be here, the shared one are here, teams are here (the 32.0.3 bug is here to https://github.com/nextcloud/circles/issues/2261 )
- Talk app: all the conversations seems to be here
- Photos app: does the job
- Collectives app: no regression so far
- Notes app: no regression so far
- Dashboard;: no regression so far
- Team Folder: no regression so far

Do you need some other kind of repport ?

Issue #854, opened 2026-01-21
Notify Push configuration is broken


p4p4j0hn:
The notify push configuration provided in the ynh package does not match the Github readme for notify push. The readme says to use a localhost port, not a unix socket. https://github.com/nextcloud/notify_push?tab=readme-ov-file#systemd

After manually configuring the systemd service file and the nginx conf file to use a local port, the self check occ command indicates everything is working.

luc38090:
I tried on my testing server this upgrade.
It seems to have worked, everything seems to be ok but I share this warnings:

```
Info: [##########+++.......] > Reapplying file permissions...
Info: [#############+++....] > Regenerating system configurations for nextcloud__2...
Info: The service nginx has correctly executed the action reload-or-restart.
Info: The service fail2ban has correctly executed the action reload-or-restart.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
Warning:   Command "/var/www/nextcloud__2/cron.php" is not defined.
[...]

PR #856, opened 2026-01-27
Notify push

ericgaspar:
- #854


ericgaspar:
doesn't seem to work...

PR #900, opened 2026-07-28
Fix acl restore

### LINKED ISSUE #899: Stop changing permissions on `$HOME/.ssh`
https://github.com/YunoHost-Apps/nextcloud_ynh/blob/bcba5c3e41f1641c982d0240a72fede6b763bb7a/scripts/upgrade#L320

This locked me out of my server twice: this line changes the home directory permissions and its subfolders, including `.ssh/`

`$HOME/.ssh` has strict permissions requirements. With this change, I cannot login anymore after nextcloud updates, and I must connect a keyboard & a screen to my server.

I understand the use-case for this option (though I do not recall ever enabling it), but I think that it should at least exclude `$HOME/.ssh`, or restore previous/stock permissions on that folder.

MayeulC:
## Problem

- Closes #899. I can target master with this PR if you prefer.

## Solution

- Check for `$user_home` before restoring acls

Note that I do not yet change permissions on .ssh. I think this is the more important fix. ~~I will make another PR to address that other case.~~ Update: added a commit that removes ACLs on .ssh specifically. A better approach may be to backup the previous value, but this will fix previously changed values, and I doubt that there is a valid use-case for other ACLs on that folder.

## PR Status

- [X] Code finished and ready to be reviewed/tested
- [X] The fix/enhancement were manually tested (if applicable)

## Automatic tests

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -151,6 +151,8 @@ then
     # Iterate over users to extend their home folder permissions
     for u in $(ynh_user_list); do
         setfacl --recursive --modify g:$app:rwX,d:g:$app:rwX "/home/$u" || true
+        # Avoid leaving acl on .ssh, which could prevent users from connecting
+        setfacl --recursive -bn "/home/$u/.ssh" || true
     done
 fi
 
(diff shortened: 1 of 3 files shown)


MayeulC:
!testme

